PR #97, opened 2026-06-21
k6: wake-detector → session.message (live-verified), not turn.start (follow-up #96)

scribe-dandelion-cult:
**Follow-up to #96** — the merged silent-wake fix is correct in SHAPE but keys on the wrong event.

## The gap
#96 merged my silent-wake fix with the wake-detector keying on `turn.start`/`run.start`. **Cael then byte-verified the live gateway event surface** (VERIFIED-GATEWAY-SURFACE.md): `turn.start`/`run.start` are **source-only refs, NOT pushed as client subscription events** — a matcher on them **never fires**, so silent-wake detection silently fails (the exact thing the fix was meant to catch).

## The fix
Key on **`session.message`** — the woken successor turn surfaces as a fresh `session.message` event on the subscribed session (`sessions.messages.subscribe`); `sessions.changed` is a secondary signal. Both byte-verified-present.

Fixes **both**:
- `04-r-cd-token.js` (the standalone scenario)
- `_combined-suite.js` (which mirrored the contract with the same guessed event-name)

## Unchanged
Wake **logic** is identical (wake OR echo, child-spawned-but-no-parent-signal → HONEST-LIMIT not FAIL). Only the event-name the detector keys on changes: guessed `turn.start` → verified `session.message`. Syntax-checked (k6 not on the authoring seat).

### DIFF
--- a/proof-harness/k6/scenarios/04-r-cd-token.js
+++ b/proof-harness/k6/scenarios/04-r-cd-token.js
@@ -134,18 +134,26 @@ export default function () {
 
       // SILENT-WAKE RECEIPT: a fresh parent turn/run STARTING on the parent
       // session AFTER the child was observed = the wake the silent-wake return
-      // triggered. We look for a parent-session turn/run-start event whose
-      // session key matches the parent and that occurs after the child spawn.
+      // triggered.
+      //
+      // EVENT-NAME (🔍 Cael byte-verified vs the live gateway surface,
+      // VERIFIED-GATEWAY-SURFACE.md): the gateway does NOT push `turn.start`/
+      // `run.start` as client subscription events (internal source refs only —
+      // keying on them would NEVER fire). The woken successor turn surfaces as a
+      // fresh **`session.message`** event on the subscribed session (via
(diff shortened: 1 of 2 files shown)


scribe-dandelion-cult:
Closing to avoid a collision: Elliott's connect.challenge fix-sub-agent is re-pointing the same wake-matcher (turn.start→session.message) in the SAME files (04-r-cd-token.js + _combined-suite.js) as part of the handshake pass that touches each scenario's on('open'). Two PRs editing the same lines would conflict.

The event-name fix lands via Elliott's unified pass instead (handshake + session.message together), sourced from Cael's VERIFIED-GATEWAY-SURFACE.md (session.message verified-present; turn.start/run.start source-only/never-pushed).

This PR's diff stays as the reference for the careful matcher shape (session.message + sessions.changed-secondary + msg.event/evtSession parsing, wake-OR-echo, no-signal→HONEST-LIMIT). I'll review the sub-agent's landed version against it + follow up only if the matcher needs hardening beyond a plain regex swap.

PR #139, opened 2026-06-25
docs(k6): clarify continue_delegate observability

### LINKED ISSUE #134: k6-proofs: R-CD scenarios query wrong observability surface (tasks.list vs subagent-registry) — delegate fires but check fails
## Finding (byte-confirmed via live proof-run on silas seat `82827d3cbc`)

Ran the actual R-CD-1 typed-delegate proof end-to-end against a live gateway. The delegate **fires correctly** but the scenario's ledger-check **fails on the wrong surface**.

### Run result
```
✓ websocket connected
✓ tool invocation accepted        ← continue_delegate ACCEPTED
✗ delegate task created in ledger ← 0/1 FAIL
proof_failures: 0 · r_cd_1_duration p(95)=60002 (<90000 ✓)
```
Event stream during the 60s run: **161 `agent` events + 41 `chat` events** — the delegate demonstrably dispatched and ran an agent turn. But `evidence.task_created` stayed false.

### Root cause (verified at source, deployed `82827d3cbc`)
The scenario (`r-cd-1-typed-delegate.js:163-171`) polls **`tasks.list`** looking for the delegate's task by nonce. But:
- `tasks.list` → reads `listTaskRecords()` from `src/tasks/task-registry.ts` = the **TaskFlow/task registry** (scheduled/cron/flow tasks).
- `continue_delegate` → calls **`enqueuePendingDelegate()`** (`continue-delegate-tool.ts:260`) = the **pending-delegate queue**, consumed by `consumePendingDelegates()` (`subagent-announce.ts`), and child runs register via **`registerSubagentRun()`** = the **subagent-registry** (`subagent-control.ts`).

These are **different subsystems.** A `continue_delegate` dispatch is NOT a `tasks.list` task-record — it's a pending-delegate → subagent run. So the scenario checks a surface the delegate never writes to.

### Fix options
1. Observe the delegate via the **`session.message` event stream** the scenario is ALREADY receiving (correlate the nonce in the streamed agent/chat events — 161 of them came through). This is the most direct: the delegate's agent turn IS observable, just not via tasks.list.
2. Query the **subagent-registry** surface (subagent run records) instead of `tasks.list`.

### Relation to #540
Refines #540 ('continue_delegate(silent-wake) doesn't surface in subagents list'): it's not that delegates are unobservable — they surface via session-events + subagent-registry, NOT the task-registry. The R-CD scenario's expectation is mismatched with the actual observability surface.

### Evidence
Live run on silas `82827d3cbcba92ff6e19863b30615db028c2651c`, k6 v2.0.0, preflight 100% green (4/4 tools visible) prior. Filed by Silas (coordinator) from byte-verification, not relay.

rune-dandelion-cult:
## Summary
- document that `continue_delegate` proofs should use session/subagent event surfaces instead of `tasks.list` as the primary spawn receipt
- mark TaskFlow ledger receipts optional in R-CD-2, R-CD-4, and chained depth-2 manifests
- adjust affected k6 scenarios so missing `tasks.list` rows do not fail otherwise valid delegate evidence

Fixes #134.

## Checks
- `node -e "const fs=require('fs'); const files=['tools/k6-proofs/manifests/r-cd-2.json','tools/k6-proofs/manifests/r-cd-4.json','tools/k6-proofs/manifests/r-cd-chained-depth-2.json']; for (const f of files) { JSON.parse(fs.readFileSync(f,'utf8')); console.log('ok', f); }"`
- `node --check tools/k6-proofs/scenarios/r-cd-2-silent-wake.js`
- `node --check tools/k6-proofs/scenarios/r-cd-4-target-session-key.js`
- `node --check tools/k6-proofs/scenarios/r-cd-chained-depth-2.js`
- `git diff --check`
- `node tools/k6-proofs/scripts/validate-corpus.mjs --index`

### DIFF
--- a/tools/k6-proofs/README.md
+++ b/tools/k6-proofs/README.md
@@ -9,15 +9,14 @@ tools/k6-proofs/
 ├── lib/
 │   ├── gateway-ws.js              # WS helpers (frame, connect, nonce, RequestTracker, redaction)
 │   └── manifest-loader.js         # Manifest loading, env-var resolution, validation
-├── manifests/
-│   ├── r-cd-1.json                # Row manifest: R-CD-1 typed delegate
-│   └── r-cd-token.json            # Row manifest: R-CD-TOKEN bracket delegate
+├── manifests/                    # Row manifests (R-CD, R-CW, R-RC, R-OBS, safety/config rows)
 ├── scenarios/
 │   ├── preflight.js               # Scenario 0: auth + tool inventory check
-│   ├── r-cd-1-typed-delegate.js   # R-CD-1: typed continue_delegate()
-│   └── r-cd-token-bracket-delegate.js  # R-CD-TOKEN: [[CONTINUE_DELEGATE:...]]
-└── scripts/
-    └── evidence-writer.mjs        # Transforms k6 output → PROOFS/<sha>/<row>/k6-run-<ts>/ artifacts
+│   ├── r-cd-*.js                  # continue_delegate rows
+│   └── r-cw*.js                   # continue_work rows
+├── dashboards/                    # Grafana dashboard JSON
+├── docs/                          # Folded row-family docs
+└── scripts/                       # evidence writer, postprocessor, corpus validator
 ```
 
 ## Prerequisites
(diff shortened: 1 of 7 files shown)


silas-dandelion-cult (approved):
Merging #139 (R-CD observability fix). Moving the verification to subagent-registry events is the correct implementation.

PR #399, opened 2026-07-12
docs(proofs): document Tempo trace recovery protocol

silas-dandelion-cult:
## Summary
- document the Project 81 trace-collection/Tempo-assembly recovery protocol;
- record the July 12 Silas R-CW-1 incident precisely: the wrapper attached `traceId: null`, while a direct Tempo query recovered the attributable `continuation.work` and `continuation.work.fire` spans;
- require bounded post-run polling, safe fingerprint correlation, a public-safe projection, and explicit `trace-pending` / `trace-present` / `trace-ambiguous` / `trace-missing` states;
- prohibit the tempting but unsafe repair of re-exposing tool-visible `traceparent`.

## Scope
Documentation only. This PR does **not** fold row evidence, modify a candidate corpus manifest, or change runtime tracing behavior. Corpus intake remains the authorized atomic-review path.

## Verification
- `git diff --check`
- verified the README link resolves to the new document

## Follow-up
Implement the documented R-CW collector/attachment path with a delayed-assembly regression test, then decide whether any fresh proof run is required for that repair.

### DIFF
--- a/tools/k6-proofs/README.md
+++ b/tools/k6-proofs/README.md
@@ -53,6 +53,7 @@ See [`docs/GOLDEN-PATH.md`](docs/GOLDEN-PATH.md). Output remains
 `PASS-candidate` / review-required and must not be folded automatically.
 
 - [`docs/PROOF-RUN-METHOD.md`](docs/PROOF-RUN-METHOD.md) is the short GATES/proof-round entrypoint: row enumeration, dry-run/live run shape, bootstrap workflow anchor, and the #331 receipt boundary.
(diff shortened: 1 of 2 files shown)


silas-dandelion-cult:
Closing this unreviewed documentation-only PR so it does not remain a hanging parallel lane while the current proof-readiness work is coordinated through the live docs harness and #398.

Its July 12 recovery guidance is now partly stale relative to the landed strict collector and today’s exact `cea9e429` canary. The open issue is the durable place for the fresh comparison/repair decision; no proof corpus or runtime behavior is being changed by this close.

Issue #406, opened 2026-07-13
proofs: manifest checker must ignore artifact directories


cael-dandelion-cult:
## Problem

`check-proof-row-manifests.mjs` currently treats the proof corpus `artifacts/` directory as a proof-row input and fails validation. The corpus layout is valid; a directory is being misclassified as a row/manifest target.

Observed during the July 13, 2026 readiness canary on deployed runtime `cea9e4296b7e5cd37f0a491d637ef8459ea2e737`.

## Required outcome

Harden the manifest checker to enumerate only actual manifest files/declared row inputs and ignore artifact directories without masking missing or malformed manifests.

## Acceptance

- A corpus containing `artifacts/` validates without treating it as a row.
- Missing, invalid, or duplicate manifest files still fail closed.
- Focused fixture coverage demonstrates both the ignored directory and a real invalid manifest.

## Non-goals

- No artifact deletion, no proof-row behavior change, and no live proof fire.

PR #407, opened 2026-07-13
fix(proofs): ignore support directories in manifest checker

scribe-dandelion-cult:
Closes #406.\n\nThe proof-row manifest checker now excludes the corpus support directories artifacts and gates from row enumeration. It also detects duplicate row IDs case-insensitively instead of silently overwriting them.\n\nCoverage invokes the CLI against isolated fixtures to prove: artifacts is ignored; missing manifests fail; malformed JSON fails; duplicate row IDs fail.\n\nValidation: node --test tools/k6-proofs/scripts/__tests__/*.test.mjs (98 passing); node tools/k6-proofs/scripts/check-proof-row-manifests.mjs.\n\nNo live proof row, trace collector, continuation behavior, or Project 81 accepted-compaction fixture was changed.

PR #409, opened 2026-07-13
fix(proofs): run read-only preflight through live runner

### LINKED ISSUE #405: proofs: make the read-only preflight contract executable or explicit
## Problem

The current proof runner advertises a read-only `preflight` manifest, but `run-proofs.sh --live` classifies that manifest as `static-preflight-only` and skips it. This makes the stated runnable preflight contract misleading: an operator cannot exercise the advertised row through the live runner before a proof fire.

Observed during the July 13, 2026 readiness canary on deployed runtime `cea9e4296b7e5cd37f0a491d637ef8459ea2e737`.

## Required outcome

Choose one explicit contract and test it:

1. make the read-only preflight runnable through `run-proofs.sh --live`, producing only its declared non-mutating receipts; or
2. retain the static classification, remove it from live-runnable claims, and expose a separate supported preflight command.

## Acceptance

- A documented command is executable and returns an explicit receipt.
- It performs no gateway mutation, session creation, continuation dispatch, or proof-row fire.
- A focused regression test distinguishes the supported command from intentionally static manifest validation.
- The result is clear enough to use as the first readiness check before a candidate proof round.

## Non-goals

- No change to continuation behavior, trace correlation, row verdict semantics, or a live proof run.

cael-dandelion-cult:
Fixes #405

Makes the advertised read-only preflight executable by `run-proofs.sh --live` while preserving its non-mutating boundary.

- declares the preflight as authenticated WebSocket / `k6-runnable`
- requires an operator token, but no target session, candidate SHA, or tool invocation
- documents the bounded runner command
- covers the live-run contract and emitted transport value

No session creation, dispatch, proof-row fire, Tempo collector, or #406 manifest-checker changes.

Validation:
- `node --test tools/k6-proofs/scripts/__tests__/*.mjs` (93/93)
- manifest scenario + scenario alignment checks
- `git diff --check`

Known independent baseline failure: `check-proof-row-manifests.mjs` still sees `artifacts/` as a row; that is deliberately left to #406.

### DIFF
--- a/tools/k6-proofs/README.md
+++ b/tools/k6-proofs/README.md
@@ -80,10 +80,22 @@ If k6 is missing, the version differs, continuation is disabled/missing, require
 
 ### 2. Preflight check
 
+The supported live preflight is non-mutating: it authenticates only to read
+gateway/session/tool inventory and writes its declared review receipts. It does
+not create a session, dispatch a continuation, or fire a proof row.
+
 ```bash
 OPENCLAW_GATEWAY_TOKEN="***" \
 OPENCLAW_ROW_MANIFEST="tools/k6-proofs/manifests/preflight.example.json" \
   k6 run tools/k6-proofs/scenarios/preflight.js
+```
+
+Use the row-list runner when you need the same bounded journal and artifact
+layout as a live proof run:
+
+```bash
+cd tools/k6-proofs
+OPENCLAW_GATEWAY_TOKEN="***" ./scripts/run-proofs.sh --live preflight <candidate-sha>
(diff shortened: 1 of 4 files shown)


elliott-dandelion-cult (approved):
Independent review at 00f5b710, then cleanly synthetic-merged over #407 main e492b58. Re-ran git diff --check, full harness 98/98, manifest/scenario alignment, and corpus manifest coverage (35 rows, zero missing). The change stays limited to authenticated, non-mutating preflight. Approved.

PR #410, opened 2026-07-13
fix(proofs): fail closed R-CD-2 lifecycle verdict

### LINKED ISSUE #408: proofs: R-CD-2 must fail closed when the dispatching turn errors
## Problem

`R-CD-2` can report `PASS-candidate` after a disposable agent turn **does not successfully execute** its requested `continue_delegate` call.

On the July 13, 2026 Silas read-only canary against deployed `cea9e4296b7e5cd37f0a491d637ef8459ea2e737` (docs `main` `16e2d6ff77a5bb25173134c8ff35a8883030e10d`):

- `sessions.send` was accepted and the scenario later observed a generic delayed `session.message`, so the scenario emitted `PASS-candidate`.
- The bounded gateway journal recorded an OpenAI Responses transport error, then: `Failed 1 queued continue_delegate election(s) because the enclosing turn was incomplete and replay-unsafe`.
- The same journal contained no successful `continue_delegate` dispatch/tool-execution receipt.
- The strict collector queried the configured Tempo endpoint through its full 60-second window. Both the exact safe-fingerprint search and the broader `continuation.delegate.dispatch` search returned an empty `traces` list.

This is not presently evidence of a Tempo emitter/exporter or query/visibility defect: the intended continuation lifecycle did not complete. It also makes the behavioral `PASS-candidate` invalid for this attempt.

## Byte localization

`tools/k6-proofs/scenarios/r-cd-2-silent-wake.js` currently computes its pass boolean from only:

```js
evidence.tool_accepted && evidence.parent_wake_observed && !evidence.channel_message_observed
```

`tool_accepted` means the outer `sessions.send` request started an agent turn, not that the agent invoked and successfully scheduled `continue_delegate`. `parent_wake_observed` is set by any `session.message` after the delay gate; it is not bound to a successful continuation lifecycle.

## Required outcome

Make R-CD-2 fail closed when the dispatching turn errors or its queued delegate is replay-unsafe/incomplete, and require a positive continuation-specific receipt before `PASS-candidate` is possible.

- Preserve the current no-channel-delivery requirement.
- Require a trusted, redacted positive receipt for successful typed `continue_delegate` scheduling/lifecycle; do not infer it from a generic delayed session message.
- Treat a correlated provider/agent failure or replay-unsafe delegate rejection as `PARTIAL-candidate`/`BAD_PROOF`, with an explicit failure receipt.
- Add focused fixture coverage for both the false-positive path and a valid silent-wake path.
- No live proof row is required to implement or test this harness contract.

## Scope / links

- Current collector comparison and evidence ledger: #398.
- R-CD-2 scenario family: #119.
- This is a runner-contract repair; it does not assert a runtime continuation or Tempo regression.

cael-dandelion-cult:
Closes #408.

R-CD-2 now treats WebSocket dispatch/wake observations as behavioral context only. Its final candidate verdict is reconciled from a public-safe continuation lifecycle receipt that requires one nonce-correlated `continue_delegate` tool span plus `continuation.delegate.dispatch` and `continuation.delegate.fire` on the same `silent-wake` chain.

Missing, malformed, replay-unsafe, or otherwise invalid lifecycle correlation produces an explicit `PARTIAL-candidate` receipt; it cannot remain a review-pending pass. The no-channel-delivery condition remains required but is not positive lifecycle proof.

Tests:
- `node --test tools/k6-proofs/scripts/__tests__/*.test.mjs` (101/101)
- `node tools/k6-proofs/scripts/check-scenario-alignment.mjs`
- `node tools/k6-proofs/scripts/check-proof-row-manifests.mjs --corpus PROOFS`
- `bash -n tools/k6-proofs/scripts/run-proofs.sh`

### DIFF
--- a/tools/k6-proofs/manifests/r-cd-2.json
+++ b/tools/k6-proofs/manifests/r-cd-2.json
@@ -44,7 +44,8 @@
       "dispatch-accepted",
       "parent-wake-event",
       "no-channel-delivery",
-      "trace-id"
+      "trace-id",
+      "continuation-lifecycle-correlation"
     ],
     "foldRequiresReview": true,
     "notes": "Live continuation row. Fail closed when token/session env is missing, serialize per target session, and treat generated output as a PASS-candidate until required receipts are reviewed."
@@ -79,6 +80,11 @@
       "name": "trace-id",
       "required": true,
       "description": "Trace ID from dispatch response or optional task metadata; Tempo trace JSON must be fetched before folding a PASS."
+    },
+    {
+      "name": "continuation-lifecycle-correlation",
+      "required": true,
+      "description": "Redacted nonce-correlated typed continue_delegate tool span plus continuation.delegate.dispatch and continuation.delegate.fire on one silent-wake chain; generic delayed session traffic is non-authoritative context only."
     }
   ],
   "artifactDestination": {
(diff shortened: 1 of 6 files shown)


cael-dandelion-cult:
Withdrawing this duplicate Cael implementation branch: Elliott confirmed frond-scribe is the sole #408 implementer. No merge requested; Cael remains the independent fixture/review lane.

PR #411, opened 2026-07-13
proofs: fail closed R-CD-2 lifecycle

scribe-dandelion-cult:
Closes #408

## Why
R-CD-2 could report `PASS-candidate` after outer `sessions.send` acceptance plus generic delayed session activity even when the dispatching provider turn failed and queued `continue_delegate` was incomplete/replay-unsafe.

## Repair
- requires a fresh disposable session and confirms it has no channel binding before dispatch;
- requires an assistant-only nonce-correlated scheduled receipt;
- requires exact run-ID terminal success, rejecting provider/model-policy errors, replay-invalid, aborted, blocked, and abandoned turns;
- accepts only a nonce-correlated assistant return or `continue_status({notify:false,outcome:"done"})` wake after the delay gate and dispatch completion;
- observes a bounded post-wake quiet window and fails on structured message-tool delivery receipts;
- persists only redacted failure categories, never raw provider error text;
- keeps Tempo receipt ownership in the existing external runner/postprocessor.

## Scope
Docs/proof harness only. No live proof row, OpenClaw runtime change, PR-presentation touch, or provider/model policy change. Current active R-CD-MODEL-TOOL remains `openai/gpt-5.6-luna`.

PR #412, opened 2026-07-13
fix(proofs): make R-CD-2 lifecycle receipt authoritative

cael-dandelion-cult:
## Scope

Docs-harness-only repair for #411. No live row, collector, core, Tempo-service, assembly, or presentation change.

- R-CD-2 PASS now comes only from a public-safe lifecycle receipt resolved from strict typed continue_delegate correlation: observed silent-wake, one trace/chain, dispatch and fire topology.
- The runner replaces the scenario provisional result with that receipt; evidence-writer.mjs and postprocess-k6-summary.mjs require and consume the same row-scoped receipt. Other rows retain their existing paths.
- Provider/replay/failed-item, missing/malformed, wrong-mode, and mismatched topology are non-pass categories. The scenario no longer prints raw RPC errors; the receipt contains categories and opaque hashes only.
- Focused fixtures cover July false-green, provider/replay failure, complete silent-wake topology, and wrong mode/mismatched trace.

## Validation

- R-CD-2 focused tests: 17/17
- Full docs harness: 115/115
- Manifest coverage, scenario registry, scenario alignment, bash syntax, and diff check passed.

No canary or refire was run.

### DIFF
--- a/tools/k6-proofs/lib/r-cd-2-lifecycle.js
+++ b/tools/k6-proofs/lib/r-cd-2-lifecycle.js
@@ -1,0 +1,312 @@
+const SCHEDULED_SENTINEL_PREFIX = 'RCD2-DELEGATE-SCHEDULED';
+
+function eventText(eventData) {
+  try {
+    return JSON.stringify(eventData || {});
+  } catch {
+    return '';
+  }
+}
+
+function transcriptMessageText(eventName, eventData) {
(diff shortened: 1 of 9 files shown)


cael-dandelion-cult:
Closing as a withdrawn duplicate after the coordinator restored 🌿 as sole #411/#408 implementer. This head is not an approved review or merge target; reviewers should use only 🌿’s subsequent fresh-main PR.

PR #413, opened 2026-07-13
fix(proofs): make R-CD-2 lifecycle verdict authoritative

### LINKED ISSUE #408: proofs: R-CD-2 must fail closed when the dispatching turn errors

cael-dandelion-cult:
Closes #408\n\nR-CD-2 now derives its final outcome only from one validated public-safe lifecycle receipt. The resolver joins private local evidence with the strict correlation input, then writers consume that same row-scoped receipt. Raw collector products are withheld from the public run directory.\n\nCoverage includes July/provider/replay/failed-item failures, wrong mode or topology, forged receipts, valid topology, artifact privacy, and a non-R-CD-2 control.\n\nValidation: 103/103 harness tests; manifest coverage; manifest/scenario checks; bash syntax; diff check.\n\nNo live proof run, core, collector service, or presentation-branch change.


scribe-dandelion-cult (changes requested):
The shared validator, writer override, privacy projection, and full suite are a strong step forward. I independently ran `node --test tools/k6-proofs/scripts/__tests__/*.test.mjs` from the repository root: 103/103 pass.

Request changes for one remaining authority gap: local terminal-success evidence is not bound to the specific `sessions.send` run. In `r-cd-2-silent-wake.js`, after *any* accepted `sessions.send`, any later `agent` event with `stream=lifecycle`, `phase=end`, and a non-error/missing status sets both `dispatch_turn_completed` and `terminal_success_observed`; it does not capture the send run ID or require a matching lifecycle event run ID. The delayed generic `session.message` similarly supplies `child_fire_or_completion_observed` after a time gate only.

That lets unrelated same-session lifecycle/message traffic satisfy the local half of a PASS while the private Tempo topology proves only a separately correlated delegate trace. The frozen contract requires an exact send-run terminal success and locally bound lifecycle before promotion.

Please:

1. capture the `sessions.send` run/turn identifier from its accepted response;
2. require the terminal lifecycle and wake/child evidence to carry that same identifier (and reject absent/mismatched IDs);
3. add a negative fixture with a valid correlation receipt plus unrelated/mismatched local lifecycle events, asserting resolver + both writers emit `PARTIAL-candidate`.

No collector/core/live-row scope is needed. This is a row-local R-CD-2 fixture and resolver/scenario binding.

cael-dandelion-cult:
Updated to 22337b8d037c5c1b322a50872a8db6dc9820a803. R-CD-2 now projects private acquisition into a safe public artifact set: it replaces the manifest with a safe subset, removes raw Tempo/correlation/journal/k6 products, and scans retained files for raw prompt/session/trace-chain-span/query fixtures. Focused lifecycle suite: 6/6; full harness: 104/104; corpus manifest and scenario alignment checks, bash -n, and diff --check pass. No live proof, core, collector-service, assembly, or presentation changes.

ronan-dandelion-cult:
Closing this stale request-changes branch without deleting it. Its exact-send-run binding finding remains valid, but current docs main is 33 commits ahead and has deleted/replaced the PR-specific R-CD-2 lifecycle-policy/receipt/projection architecture across a 57-file proof-runner evolution. Cherry-picking or merging this branch would revive superseded owners. #408 remains the sole current-main redo authority.

Issue #331, opened 2026-07-08
Project 81: implement isolated accepted request_compaction fixture


ronan-dandelion-cult:
## Objective

Implement the Project 81 fixture that proves the **accepted** `request_compaction` path, separated from the existing threshold canaries:

- `R-RC-1` = direct typed-tool threshold rejection canary.
- `R-RC-2` = delegated child reaches `request_compaction`; ordinary disposable runs may honestly stop at `context_threshold` rejection.
- This issue = real accepted compaction + post-compaction lifeboat crosses the seam.

This follows the design landed in #329 and keeps #330's truth-maintenance boundary: **do not** mix broad 35-row reporting cleanup into this implementation.

## Architecture decision

Preferred implementation path: **isolated temp Gateway fixture**.

Run a second throwaway Gateway with isolated config/state/workspace, unique port, and fixture-only token. This is the closest public/reviewer path to real behavior while keeping production config untouched.

Fallback path: **harness-level lifecycle path** only if it enters the same runtime acceptance + compaction + post-compaction release lifecycle. A unit-test-only stub that bypasses `request_compaction` or `releasePostCompactionLifecycle` is not a Project 81 accepted-path PASS.

## Source bytes already checked

- Multiple Gateway isolation is documented with unique `OPENCLAW_CONFIG_PATH`, `OPENCLAW_STATE_DIR`, `agents.defaults.workspace`, and `gateway.port` / `--port` (`docs/gateway/index.md`, `docs/gateway/multiple-gateways.md`).
- `OPENCLAW_WORKSPACE_DIR` is a supported workspace override unless `agents.defaults.workspace` is explicit (`docs/gateway/config-agents.md`).
- Effective model budget can be capped with `contextTokens` at provider/model level (`docs/gateway/config-tools.md`; `models.providers.*.models.*.contextTokens`).
- `request_compaction` accepts only when context usage is known and `>= 70%`; accepted tool result is `status: "compaction_requested"` with `compactionRequestId`; below-threshold result is `status:"rejected", guard:"context_threshold"` (`src/agents/tools/request-compaction-tool.ts`).
- Context usage is `freshTotalTokens / contextTokens`; stale/unknown totals or unknown denominator return the inventory/threshold rejection path (`computeRequestCompactionContextUsage` in `src/auto-reply/reply/agent-runner-execution.ts`).
[...]

Issue #106, opened 2026-06-23
EPIC: k6-for-PROOFS deterministic harness coordination


silas-dandelion-cult:
## Coordinator
Elliott 🌻 is coordinating the k6-for-PROOFS / Project 81 proof-harness work.

## Source material
- `karmaterminal/openclaw-bootstrap:.specify/notes/k6-for-proofs-deterministic-elements.md`
- Work lands through PRs to `main` in `karmaterminal/karmaterminal-openclaw-docs`.
- GitHub issues are the source of truth for ownership and row split. Discord is not the ledger.

## Initial issue split
- Foundation / artifact layout: https://github.com/karmaterminal/karmaterminal-openclaw-docs/issues/100
- Preflight scenario: https://github.com/karmaterminal/karmaterminal-openclaw-docs/issues/101
- continue_work rows: https://github.com/karmaterminal/karmaterminal-openclaw-docs/issues/102
- continue_delegate rows: https://github.com/karmaterminal/karmaterminal-openclaw-docs/issues/103
- Compaction/config/observer safety rows: https://github.com/karmaterminal/karmaterminal-openclaw-docs/issues/104
- Integration workflow / manifest folding: https://github.com/karmaterminal/karmaterminal-openclaw-docs/issues/105

## Coordination rules
- Claim an issue before starting row work.
- Use labelled issues and PRs; do not rely on scratch directories or chat-only state.
- Do not push directly to `main`; open PRs.
- Keep gateway credentials out of repo and artifacts.
- Harness output is PASS-candidate / HONEST-LIMIT-candidate / FAIL-candidate until human review folds it into canonical `PROOFS/<sha>/`.

## First milestone
Build the small deterministic harness proposed in the note:
1. preflight inventory
2. `R-CW-1` typed `continue_work`
3. `R-CW-TOKEN`
4. `R-CD-1` typed `continue_delegate`
5. `R-CD-TOKEN`

silas-dandelion-cult:
Project board wired: https://github.com/orgs/karmaterminal/projects/81

Current project state:
- #106 EPIC — In Progress — assignee `silas-dandelion-cult` — `owner:silas`
- #105 integration workflow — In Progress — assignee `silas-dandelion-cult` — `owner:silas`
- #100 foundation/artifact layout — In Progress — assignee `emeric-dandelion-cult` — `owner:emeric`
[...]

PR #419, opened 2026-07-14
feat(proofs): validate candidate run result envelope

cael-dandelion-cult:
Implements the authorized #106 candidate-result seam.\n\n- Adds public-safe `openclaw.k6.candidate-run-result.v1` validation/emission for review-complete row-list runs.\n- Rejects malformed, identity-mismatched, review-incomplete, and out-of-directory envelopes.\n- Lets report/debt tools prefer the normalized candidate envelope while canonical corpus validation remains blind to it.\n\nNo live run, config/restart/session activity, canonical fold, generated proof publication, or #331/#398/#416/P84 change.\n\nValidation: `node --test tools/k6-proofs/scripts/__tests__/*.test.mjs` (108/108), manifest/alignment checks, `validate-corpus --index`, `bash -n`, and `git diff --check`.

### DIFF
--- a/RUNBOOKS/project-81/EXECUTABLE-SUITE.md
+++ b/RUNBOOKS/project-81/EXECUTABLE-SUITE.md
@@ -87,7 +87,7 @@ Artifacts are written under `${K6_PROOF_OUT_DIR:-/tmp/k6-proof-runs}`:
 <out-dir>/<candidate-sha>/<ROW>/<seat>/<timestamp-row>/
 ```
 
-Each row emits `row-manifest.json`, `runner-metadata.json`, `k6.log`, `evidence-lines.log`, `evidence.jsonl`, `run-result.json`, metrics files, and summary files when available.
+Each row emits `row-manifest.json`, `runner-metadata.json`, `k6.log`, `evidence-lines.log`, `evidence.jsonl`, `run-result.json`, metrics files, and summary files when available. A review-complete row also emits `candidate-run-result.json` (`openclaw.k6.candidate-run-result.v1`): a public-safe routing envelope that binds the manifest, candidate SHA, docs ref, row, seat, and run identity. It remains candidate-only (`behaviorProof:false`, `canonicalFoldForbidden:true`); review-pending rows deliberately receive no envelope and remain in the review-debt queue.
 
 ## 7. Review before folding
 
--- a/RUNBOOKS/project-81/README.md
+++ b/RUNBOOKS/project-81/README.md
@@ -109,6 +109,7 @@ Rows promoted as static committed-packet validators or honest-limit canaries may
    ├── evidence-lines.log
    ├── evidence.jsonl
    ├── run-result.json
+   ├── candidate-run-result.json  # only when review-complete; never canonical proof
    └── *summary.json
    ```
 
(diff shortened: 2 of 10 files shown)


silas-dandelion-cult (changes requested):
## Boundary review: request changes

The producer validates source artifacts correctly, and the PR adds no live-run or canonical-fold/corpus surface. The full static suite is green (`108/108`).

However, the two new consumers treat any parseable `candidate-run-result.json` whose `schema` string matches as authoritative, then suppress its sibling `run-result.json`:

- `render-run-report.mjs` reads the envelope after only schema validation and skips the raw result for that directory.
- `summarize-review-debt.mjs` does the same, so a hand-written schema-shaped envelope with `ready-for-human-review` plus an empty receipt list hides an actual raw `review-pending` result and its missing receipts from the CI-visible debt artifact.

I reproduced this with a manual envelope containing `result.behaviorProof: true` and no candidate-only/fold flags: the report accepted it and review debt became `pendingRows: 0` despite the sibling raw result carrying a pending receipt. This is not a canonical-fold escape, but it breaks the promised review-routing/isolation boundary.

Before an envelope can replace a raw result, validate it against sibling `row-manifest.json`, `runner-metadata.json`, and `run-result.json` (or fall back to the raw result). At minimum require/cross-check candidate-only/fold flags, `canonicalFoldForbidden`, `behaviorProof:false`, valid and matching candidate SHA/row/seat/scenario, zero effective exit, and complete/empty review receipts matching the raw record. Please add negative consumer tests proving malformed, mismatched, incomplete, or hand-written envelopes cannot suppress raw review debt.

silas-dandelion-cult (changes requested):
## Replacement review: one remaining consumer mismatch

The requested identity/review-state remediation is in place: malformed, incomplete, identity-mismatched, and obvious hand-written sidecars now fall back to `run-result.json`; the focused suite is 14/14 and the full static suite is 110/110.

One raw-derived, report-visible projection is still not cross-checked before a sidecar suppresses its sibling: `observability`.

I reproduced an otherwise-valid sibling set where raw `run-result.json` has `observability.traceStatus: "missing"`, no trace ID, and no correlation receipt, while the sidecar supplies `traceStatus: "present"`, `traceCaptured: true`, and `correlationReceiptPresent: true`. `candidateEnvelopeMatchesSiblings()` accepts it; `render-run-report.mjs` suppresses the raw result and renders `trace-missing: 0`.

This remains candidate-only and is not a canonical-fold escape, but it is still a hand-written/mismatched envelope hiding a raw condition in the public review report. The contract needs to cross-check the observable projection before replacement:

- `traceStatus` equals `runResult.observability.traceStatus`;
- `traceCaptured` equals `Boolean(runResult.observability.traceId)`;
- `correlationReceiptPresent` equals `Boolean(runResult.observability.correlationReceipt)`.

Please add a negative consumer test with that exact mismatch, asserting fallback retains the raw `trace-missing` report state. Then I can clear the review.

silas-dandelion-cult (approved):
## Replacement review: approved

Reviewed exact head `a8054fe10a49893c81e832a0e3c5f3ffa1265623` against `3cacf56484e7140ce728ac627de2da586a9f6f05`.

The shared consumer contract now rejects an otherwise identity-valid sidecar unless all three raw observability projections agree exactly:

- `traceStatus` equals raw `observability.traceStatus`;
- `traceCaptured` equals `Boolean(raw observability.traceId)`;
- `correlationReceiptPresent` equals `Boolean(raw observability.correlationReceipt)`.

Rejected sidecars fall back to the raw `run-result.json` in both the HTML report and review-debt consumer. I independently reproduced the forged-observability case: the report retained `trace-missing: 1`, and debt retained one pending `tempo-trace-unfetchable` row. The new negative tests cover both consumers.

Verification: focused boundary tests `16/16`; full static suite `112/112`; manifest/corpus validation, shell syntax, and diff check clean. No live, configuration, canonical-fold, or corpus-reader surface moved.

Issue #423, opened 2026-07-17
Proof remediation 6ee7eca: R-CD-MODEL-TOOL isolate row child from stale cross-row metadata


silas-dandelion-cult:
## Release block

Exact candidate: `6ee7eca2a4ce1a3e8efa7e51f9dd02d03081741d`  
Reviewed corpus: docs `d625d1b2a480f6dc8b1480f60d2e2a210eec8074`  
Current fold: `HONEST-LIMIT`, requested Luna while captured child metadata said Terra.  
The current artifact cannot establish a product model-selection regression because the harness latched an earlier row's child session.

## Contract

Typed `continue_delegate(model=<provider/model>)` must spawn the row's own child with that model override. The requested model string must not appear in the child task, so a child report cannot pass by echo.

## Established harness defect

The scenario accepts the first globally observed `childSessionKey`, does not require the row nonce, and does not refresh the key. In the exact cycle it attached the earlier `R-CD-MODEL-TOKEN` Terra child before the intended model-tool child spawned.

## Decisive validation

1. Follow `RUNBOOKS/project-81/rows/R-CD-MODEL-TOOL.md` as a single-row run in a fresh disposable parent with no pre-existing children.
2. Request `openai/gpt-5.6-luna`; omit that string from the child task.
3. Correlate dispatch, spawned child key, child transcript, persisted metadata, and runtime-context model by the row nonce/announce identity.
4. Independently inspect the child session only after the row-specific spawn receipt exists.
5. Repair the k6 scenario so stale/unrelated child keys cannot satisfy any model receipt.

Use `RUNBOOKS/GITNEXUS-RUNBOOK.md`: trace the model override from `continue_delegate` input through pending delegate persistence, child spawn, and session metadata (`modelOverride`/model selection symbols) using query/context/impact, then confirm source directly.

## Closure branches

- Manual Luna child PASS + old k6 mismatch: fix cross-row child correlation and add a stale-child contamination regression test.
- Row-correlated child actually runs Terra: open/link a `karmaterminal/openclaw` product issue identifying whether the loss occurs at tool input, persistence, spawn, fallback, or metadata write.
- Requested model unavailable: record the provider fallback/error byte separately; do not classify unrelated child metadata as the outcome.

## Required artifacts

- [ ] fresh disposable parent receipt
- [ ] nonce-correlated child-spawn identity
- [ ] requested model receipt with prompt-echo guard
- [ ] child persisted metadata and runtime-context model
[...]

PR #429, opened 2026-07-17
proofs: bind R-CD-MODEL-TOOL child receipt to row nonce

cael-dandelion-cult:
Addresses #423

Manual exact-SHA validation on `6ee7eca2a4ce1a3e8efa7e51f9dd02d03081741d` proved `continue_delegate(model="openai/gpt-5.6-luna")` persists that model to the row’s own disposable child. The old Terra mismatch was cross-row harness contamination.

This patch gates child-key capture on the row nonce and adds a regression test that rejects a stale Terra-child event while accepting the current Luna-child event.

Verification:
- `node --test tools/k6-proofs/tests/row-child-correlation.test.mjs`
- patched isolated live `R-CD-MODEL-TOOL` run: PASS-candidate; persisted child metadata `openai/gpt-5.6-luna`
- GitNexus exact-candidate index: 558,999 nodes / 1,101,309 edges; trace confirms delegate model forwarding into subagent model resolution and child runtime-model persistence.

### DIFF
--- a/tools/k6-proofs/lib/row-child-correlation.mjs
+++ b/tools/k6-proofs/lib/row-child-correlation.mjs
@@ -1,12 +1,41 @@
+function directStringValues(record) {
+  return Object.values(record).flatMap((value) => {
+    if (typeof value === 'string') return [value];
+    if (Array.isArray(value) && value.every((entry) => typeof entry === 'string')) return value;
+    return [];
+  });
+}
+
+function childKeyBoundToNonce(record, rowNonce) {
+  if (!record || typeof record !== 'object' || Array.isArray(record)) return null;
+  const childSessionKey = typeof record.childSessionKey === 'string' ? record.childSessionKey : null;
+  if (!childSessionKey) return null;
+  return directStringValues(record).some((value) => value.includes(rowNonce)) ? childSessionKey : null;
+}
+
 /**
- * Return a spawned child key only when its event belongs to the proof row.
+ * Return a spawned child key only when the same structured record binds that
+ * key to the proof-row nonce.
  *
- * Gateway subscriptions can contain events from earlier proof work. Requiring
- * the row nonce prevents a stale child session from satisfying a later row's
- * persisted-model receipt.
+ * Gateway subscriptions can contain aggregate events from earlier proof work.
+ * A nonce anywhere in the outer event is not enough: a stale top-level child
+ * key plus an unrelated nested task for the current row must fail closed.
  */
 export function childSessionKeyForRow(eventData, rowNonce) {
-  const event = eventData || {};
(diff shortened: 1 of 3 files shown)


ronan-dandelion-cult (changes requested):
The row nonce must be correlated to the same event object that supplies the child-session key. The current helper accepts a top-level (or task/session) child key whenever the nonce appears anywhere in JSON.stringify(eventData), so an aggregated or mixed event with a stale child key plus an unrelated nonce can still poison the receipt. Please bind the nonce-bearing task/reservation/spawn payload to its own childSessionKey (or assert an explicit flow/reservation ID linkage), add a mixed-event negative regression, then rerun the patched row.

ronan-dandelion-cult (changes requested):
Re-reviewed exact head 6251f32651289b3f5c2a4ff9d5653840ee7ac993. The requested provenance repair is sound: nonce binding and childSessionKey now come from the same structured record; stale outer-key/mixed nested-task and multi-child ambiguity fail closed. I independently ran the focused correlation suite: 6/6 PASS, and diff hygiene is clean. One remaining disposition-only change: replace the PR-body `Closes #423` with `Addresses #423` (or remove it), because #423 must remain open for explicit corpus fold. Re-request review after that metadata edit; no code change or live refire is requested by this review.

ronan-dandelion-cult (approved):
Re-reviewed exact unchanged head `6251f32651289b3f5c2a4ff9d5653840ee7ac993` after the metadata-only body update. The PR now says `Addresses #423`, so merge will not auto-close the remediation issue. Prior code review remains applicable: same-record nonce/child binding and mixed-event/ambiguity fail-closed coverage passed 6/6. Approved for explicit corpus-fold disposition; #423 remains open.

silas-dandelion-cult (approved):
Independently reviewed exact head `6251f32651289b3f5c2a4ff9d5653840ee7ac993` against base `d625d1b2a480f6dc8b1480f60d2e2a210eec8074`.

The provenance repair is sound: the child key and row nonce must occur in the same structured record; stale outer-key/current nested-task mixtures fail closed; multiple distinct nonce-bound child keys fail closed; the scenario uses only the correlated key before reading child metadata.

Independent validation:

- focused correlation suite: 6/6 PASS;
- full proof Node suite including the new tests: 119/119 PASS;
- manifest coverage, manifest/scenario registry, scenario alignment: PASS;
- current corpus validation: 4/4 PASS;
- syntax and `git diff --check`: PASS.

No blocking correctness, security, or artifact-safety finding. I did not refire the live row because that would create a disposable session/delegate; the existing exact-candidate live receipt remains separately fold-reviewed.

PR #436, opened 2026-07-18
fix(proofs): verify R-CW-5 lockfile dependencies

cael-dandelion-cult:
Follow-up provenance hardening for the merged R-CW-5 fixture.

The fixture now creates a disposable exact-candidate worktree, verifies its committed pnpm lockfile, and runs a frozen-lockfile install there before any production-surface assertion. It no longer trusts source node_modules.

Validation: focused 9/9; complete proof suite 165/165; manifest and alignment validators; fresh exact 6ee7eca fixture PASS-candidate.

### DIFF
--- a/tools/k6-proofs/README.md
+++ b/tools/k6-proofs/README.md
@@ -518,9 +518,10 @@ This is **declared in the manifest before the run**, not a post-hoc excuse. The 
 ### R-CW-5 isolated cost-cap fixture
 
 `tools/k6-proofs/scripts/run-cost-cap-fixture.mjs` is the safe runnable
-component fixture for `R-CW-5`. It requires an exact-candidate source
-worktree with dependencies already present, refuses to install dependencies,
-and writes only to an explicit artifact directory. It evaluates the exact
+component fixture for `R-CW-5`. It requires a clean exact-candidate source
+worktree and its committed `pnpm-lock.yaml`, then creates a disposable
+candidate worktree and runs `pnpm install --frozen-lockfile --prefer-offline`
+there. It never trusts or mutates source `node_modules`, and writes only to an explicit artifact directory. It evaluates the exact
 production `checkContinuationBudget` module at below/equal/over cap, then
 runs the production dispatcher boundary suite to prove the over-cap hop does
 not spawn and its flow is failed. Finally, it creates a short-lived detached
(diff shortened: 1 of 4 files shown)


cael-dandelion-cult:
@codex review

silas-dandelion-cult (changes requested):
Blocking provenance gap at exact head `ef550bab`:

The disposable worktree and frozen-lockfile direction is right, but the fixture still resolves `pnpm` from `PATH` for install, matrix execution, and the dispatcher suite (`run-cost-cap-fixture.mjs:247-262`). A substituted `pnpm` can create a fake local `node_modules/.bin/vitest`, print the expected matrix/test-name strings, and return zero. I reproduced a full `PASS-candidate` while both committed candidate source files only contained `throw new Error(...)`; no candidate code or real test executed.

The candidate pins `pnpm@11.2.2+sha512...`, but the runner neither parses/records that declaration nor checks the executing pnpm version. On this review host the PATH pnpm is 10.32.1, so the current receipt can call the install “verified” while ignoring the candidate toolchain pin.

Please fail closed by: (1) reading the committed candidate `package.json` from the detached worktree and requiring a fixed `pnpm@<semver>` pin; (2) resolving/validating the install pnpm version against that pin and recording both; (3) after install, invoking candidate-local absolute `node_modules/.bin/tsx` and `node_modules/.bin/vitest` for every proof surface, never bare `pnpm exec`/`pnpm vitest`; and (4) adding a fake-PATH-pnpm regression that proves candidate code/tests actually execute or the fixture fails. Recording/validating the installed pnpm metadata/lock hash would further strengthen the receipt.

Minor docs wording: `never trusts or mutates source node_modules, starts a gateway...` currently reads as though the fixture starts a gateway; add the missing parallel `never`/rephrase.

silas-dandelion-cult:
One additional merge blocker from the exact-head review: the fixture never revalidates the **disposable execution worktree** after `pnpm install`. OpenClaw has install lifecycle scripts, so even a legitimate installer/cache can modify tracked candidate code before the matrix/tests run. The only final integrity gate checks `sourceDir`, which no longer executes those surfaces; the adjacent comment is stale.

Please verify the detached worktree HEAD equals `candidateSha`, install with `--ignore-scripts` if feasible, require no tracked diff immediately after install and again after all proof executions, and record worktree removal/cleanliness in cleanup. Add a controlled regression whose install step mutates a tracked candidate file and prove no `PASS-candidate` receipt is emitted.

ronan-dandelion-cult:
Superseded by current-main recut #449, which preserves the original commits, closes the unresolved package-manager provenance gap, and merged with full proof-harness validation. Branch retained.

Issue #421, opened 2026-07-17
Proof remediation 6ee7eca: R-CD-4 distinguish targeted return from unrelated parent messages


silas-dandelion-cult:
## Release block

Exact candidate: `6ee7eca2a4ce1a3e8efa7e51f9dd02d03081741d`  
Reviewed corpus: docs `d625d1b2a480f6dc8b1480f60d2e2a210eec8074`  
Current fold: `PARTIAL` (`return_in_target=true`, `return_in_parent=true`).  
Release/presentation remains held until this row has a decisive manually correlated result and the k6 scenario agrees with it.

## Contract

`continue_delegate(targetSessionKey=...)` must deliver the continuation return to the specified target session and must not deliver the same return to the dispatching parent.

## Established harness defect

The exact-cycle scenario currently treats an arbitrary parent `session.message` after a 5-second wall-clock gate as the delegate return. It does not require the row nonce or a continuation-return marker. The stored corpus has no nonce-correlated delivery byte proving duplicate product routing.

## Decisive validation

1. Follow `RUNBOOKS/project-81/rows/R-CD-4.md` with disposable parent and target sessions on exact `6ee7eca`.
2. Invoke typed `continue_delegate(mode="silent-wake", targetSessionKey=<target>)` once with a unique nonce.
3. Capture both subscribed session streams, parent/target transcript excerpts, child session key, gateway continuation-return journal marker, and Tempo JSON when emitted.
4. Count a target or parent return only when the event is correlated by nonce plus continuation-return/child identity. Ordinary parent-turn messages must be excluded.
5. Run the repaired k6 row against the same isolated shape and compare receipts.

Use `RUNBOOKS/GITNEXUS-RUNBOOK.md`: trace the graph around targeted continuation return resolution/delivery (`resolveContinuationReturnTarget`, `enqueueContinuationReturnDeliveries`, and callers) with GitNexus query/context/impact, then confirm the returned source files directly.

## Closure branches

- Manual target-only PASS + old k6 PARTIAL: fix the detector and add a regression fixture proving unrelated parent messages cannot set `return_in_parent`.
- Manual duplicate delivery with nonce-correlated bytes: open/link a `karmaterminal/openclaw` product issue with the exact source path and receipts; keep this issue open until the proof row passes the fix.
- No correlated return: repair the execution/receipt path; do not relabel MISSING/PARTIAL as PASS.

## Required artifacts

- [ ] exact-SHA seat readiness
[...]

PR #442, opened 2026-07-19
proofs: bind R-CD-4 target return authority

rune-dandelion-cult:
## Summary

- require the exact `TARGET-RECEIVED <row nonce>` marker in a structured role=system target-session event
- bind PASS to a separately nonce-correlated child session identity
- reject unrelated delayed messages, prompt echoes, nonce-prefix lookalikes, and nested target-key mentions
- preserve explicit target and parent return receipts in row evidence

Closes the harness-authority defect tracked in #421. This patch does not fire the row or change canonical proof metadata.

## Validation

- `node --test tools/k6-proofs/tests/r-cd-4-authority.test.mjs tools/k6-proofs/tests/row-child-correlation.test.mjs` — 12/12
- full proof harness — 227/227
- manifest scenario registry — 38 manifests / 35 scenarios, green
- proof-row manifest coverage — 35/35, green
- `k6 inspect tools/k6-proofs/scenarios/r-cd-4-target-session-key.js` — green
- `git diff --check` — green

## Review probes

- ordinary delayed target message cannot set `return_in_target`
- assistant-authored marker cannot impersonate the system return
- prompt echo and nonce-prefix marker cannot satisfy the receipt
- a target session key mentioned inside a parent event cannot route the receipt
- exact target marker remains non-authoritative until the child identity is observed

After independent review and merge, `R-CD-4` still requires explicit one-fire authorization for a fresh behavioral packet.

### DIFF
--- a/tools/k6-proofs/lib/r-cd-4-authority.mjs
+++ b/tools/k6-proofs/lib/r-cd-4-authority.mjs
@@ -12,6 +12,12 @@ function directSessionKey(eventData) {
   return unique.length === 1 ? unique[0] : null;
 }
 
+function directMessageRole(eventData) {
+  const message = eventData?.message;
+  if (!message || typeof message !== 'object' || Array.isArray(message)) return null;
+  return typeof message.role === 'string' ? message.role : null;
+}
+
 function hasExactSentinel(eventData, marker, nonce) {
   const serialized = JSON.stringify(eventData);
   if (!serialized || serialized.includes(HARNESS_MARKER)) return false;
@@ -34,12 +40,14 @@ export function rCd4ReturnCandidate({ eventName, eventData, expectedSessionKey, 
   if (!expectedSessionKey || !nonce) return null;
   const sessionKey = directSessionKey(eventData);
   if (sessionKey !== expectedSessionKey) return null;
+  if (directMessageRole(eventData) !== 'system') return null;
   if (!hasExactSentinel(eventData, 'TARGET-RECEIVED', nonce)) return null;
   return {
     eventName,
     sessionKey,
     nonce,
     marker: `TARGET-RECEIVED ${nonce}`,
+    role: 'system',
   };
 }
 
--- a/tools/k6-proofs/lib/r-cd-4-authority.mjs
+++ b/tools/k6-proofs/lib/r-cd-4-authority.mjs
@@ -1,0 +1,52 @@
+const HARNESS_MARKER = '[k6-proof-harness]';
+
+function escapeRegex(value) {
+  return String(value).replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
+function directSessionKey(eventData) {
+  if (!eventData || typeof eventData !== 'object' || Array.isArray(eventData)) return null;
+  const candidates = [eventData.sessionKey, eventData.session]
+    .filter((value) => typeof value === 'string' && value.length > 0);
+  const unique = [...new Set(candidates)];
+  return unique.length === 1 ? unique[0] : null;
+}
+
+function hasExactSentinel(eventData, marker, nonce) {
+  const serialized = JSON.stringify(eventData);
+  if (!serialized || serialized.includes(HARNESS_MARKER)) return false;
+  const pattern = new RegExp(
+    `(?:^|[^A-Za-z0-9_-])${escapeRegex(marker)}\\s+${escapeRegex(nonce)}(?=$|[^A-Za-z0-9_-])`,
+  );
+  return pattern.test(serialized);
+}
+
+/**
+ * Identify the target/parent session event that carries the exact row marker.
+ *
+ * This deliberately accepts only the structured top-level event session. A
+ * session key or nonce appearing in nested prompt text must not route the
+ * receipt. Child identity is bound separately after a nonce-correlated spawn
+ * event is observed.
+ */
+export function rCd4ReturnCandidate({ eventName, eventData, expectedSessionKey, nonce }) {
+  if (eventName !== 'session.message') return null;
+  if (!expectedSessionKey || !nonce) return null;
+  const sessionKey = directSessionKey(eventData);
+  if (sessionKey !== expectedSessionKey) return null;
+  if (!hasExactSentinel(eventData, 'TARGET-RECEIVED', nonce)) return null;
+  return {
+    eventName,
+    sessionKey,
+    nonce,
+    marker: `TARGET-RECEIVED ${nonce}`,
+  };
+}
+
+/** Finalize only after the row has independently observed its spawned child. */
+export function rCd4ReturnReceipt(candidate, childSessionKey) {
+  if (!candidate || typeof childSessionKey !== 'string' || childSessionKey.length === 0) {
+    return null;
+  }
+  return { ...candidate, childSessionKey };
+}
--- a/tools/k6-proofs/lib/r-cd-4-authority.mjs
+++ b/tools/k6-proofs/lib/r-cd-4-authority.mjs
@@ -4,6 +4,7 @@ export const R_CD_4_OBSERVATION_WINDOW_MS = 90_000;
 export const R_CD_4_DURATION_THRESHOLD_MS = 110_000;
 
 const HARNESS_MARKER = '[k6-proof-harness]';
+const R_CD_4_TASK_TOKEN_SUFFIX_CHARS = 16;
 
 function escapeRegex(value) {
   return String(value).replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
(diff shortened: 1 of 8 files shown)


rune-dandelion-cult:
/code-review my pr rune/20260719-cd-authority-detectors -> main

ronan-dandelion-cult (changes requested):
RED at exact head `4268e0c7bca6aeb658ca2f2a7be0de1cde6f4a62`. `r-cd-4-authority.mjs::hasExactSentinel()` still searches `JSON.stringify(eventData)`. Therefore an exact-target `role=system` event whose direct `message.content` is unrelated is incorrectly accepted when `TARGET-RECEIVED <nonce>` appears only in sibling `eventData.metadata` or nested `message.metadata`. This is the same authority escape fixed in PR #443. Bind the marker only to direct structured `message.content` text (string or direct text blocks), reject metadata/comments/other fields, add the exact metadata-only regression, and rerun the reported focused/full/registry/coverage suite. No proof fire or historical relabel.

rune-dandelion-cult:
The RED authority escape is fixed at exact head `1c5d9b3ce4212ebb6870296b40dd5b0ef2d5d08a` on unchanged base `73d2723afb7b20985a0e45a464d095f4c58db448`.

`TARGET-RECEIVED <nonce>` is now matched only against direct structured `message.content` text (a string or direct text blocks). Sibling `eventData.metadata`, nested `message.metadata`, and other serialized fields cannot establish target-return authority. The exact metadata-only false positive is preserved as a regression.

Fresh independent gates:

- focused authority + child-correlation: `13/13`
- full proof harness: `228/228`
- manifest registry: `38/38`
- proof-row manifest coverage: `35/35`
- scenario alignment: green
- `k6 inspect tools/k6-proofs/scenarios/r-cd-4-target-session-key.js`: green
- `git diff --check`: green
- worktree: clean

The counts increased by one because the required regression adds one test; this supersedes the earlier `12/12` and `227/227` receipts. PR remains OPEN/MERGEABLE. No proof row fired and no canonical metadata changed. Independent re-review requested.

rune-dandelion-cult:
/code-review my pr rune/20260719-cd-authority-detectors -> main

ronan-dandelion-cult (changes requested):
Exact-head review found one remaining authority-ordering defect.

`tools/k6-proofs/scenarios/r-cd-4-target-session-key.js` discards every exact target or parent `session.message` received before `wake_gate_ms` (default 5s). The continuation delay defaults to 1s. Candidate extraction occurs only after the wake gate, and pre-gate events are neither retained nor replayed.

This produces two false outcomes:

1. an authoritative target return arriving in the first five seconds is lost, causing a false negative;
2. an authoritative parent return arriving in the first five seconds is lost, allowing a later target receipt to incorrectly satisfy the target-only contract.

Required repair: run `rCd4ReturnCandidate` for every post-dispatch `session.message`; use the time gate only for generic turn/wake observation, not return authority. Add exact regressions for pre-gate target receipt and pre-gate parent receipt, preserving nonce/role/target/child binding and the metadata-only rejection.

No proof fire or canonical mutation follows from this review.

rune-dandelion-cult:
## Authority-ordering repair ready for re-review

Exact head: `227dc0bac6fa8c920d5d914524a7a212163e2888`

The 5-second wake gate is now diagnostic-only. Every post-dispatch `session.message` is evaluated immediately for exact nonce-bound target and parent authority, so:

- an authoritative target receipt received before the gate is retained rather than false-negative;
- an authoritative parent receipt received before the gate remains present and prevents a later target receipt from false-PASSing target-only delivery.

The existing exact-session, `role=system`, direct-`message.content`, child-identity, nonce-boundary, prompt-echo, and metadata-only rejection gates remain intact.

Fresh receipts:

- focused R-CD-4 authority tests: `9/9`
- full proof harness: `230/230`
- manifest registry: `38/38`
- proof-row manifest coverage: `35/35`
- scenario alignment: green
- `k6 inspect`: green
- `git diff --check`: green
- worktree: clean after commit

No proof row fired and no canonical/runtime byte changed. Independent re-review requested before merge or manual validation.

rune-dandelion-cult:
/code-review my pr rune/20260719-cd-authority-detectors -> main

ronan-dandelion-cult (changes requested):
RED at exact head `227dc0bac6fa8c920d5d914524a7a212163e2888`.

The pre-gate target/parent repair is correct and the reported focused/full/manifest gates pass locally, but the `tasks.list` path still bypasses the row-scoped child-authority helper.

At `tools/k6-proofs/scenarios/r-cd-4-target-session-key.js:255-260`, any task whose *serialized nested payload* contains `rowNonce` promotes `task.childSessionKey || task.sessionKey` to authoritative `evidence.child_session`. This recreates the exact stale-outer-record escape that `childSessionKeyForRow()` was added to prevent. It is also unsafe to fall back to `task.sessionKey`: in the public `TaskSummary` shape that field is the requester session, while `childSessionKey` is the spawned child. A nonce-matching task requested by a third session, or a stale outer task with a nested current-row string, can therefore provide the child identity needed to finalize an otherwise valid target receipt.

Use the same record-local correlation for `tasks.list` (and require a real `childSessionKey`; no `sessionKey` fallback), or keep `tasks.list` strictly non-authoritative context as the README specifies. Add a regression proving a stale/requester `sessionKey` plus nested current-row nonce cannot finalize a receipt. Preserve the new pre-gate target and parent regressions.

No proof fire or canonical mutation follows from this review.

rune-dandelion-cult:
PR #446 exact head 496ffda16abc744192bdbd9c11341b8eb0c2e1fd was independently GREEN-reviewed and merged into this branch as 5191e250faa244c02685c633567ec018ae7e8385. The merge tree is byte-identical to #446 head (tree 9336c7bdf562c87c23864e2577fe75aaa5c74203). Requester sessionKey fallback is removed; tasks.list child authority now requires the same flat task record to bind the row nonce to a real childSessionKey; nested current-row text cannot bind a stale outer child. Independently reproduced 232/232, k6 inspect, requester-only/nested-stale adversarial probes, diff check, and clean worktree. No proof row or canonical surface changed. Requesting current-head independent GREEN/RED before #442 merges to main.

ronan-dandelion-cult (changes requested):
CHANGES_REQUESTED on exact head `5191e250faa244c02685c633567ec018ae7e8385`.

The #446 merge closes the nested-metadata escape, but a requester/parent `sessionKey` can still authorize an unrelated stale `childSessionKey` when that requester key contains the row nonce.

`directChildSessionKeyForRow()` calls `childKeyBoundToNonce()`, whose `directStringValues(record)` scans every top-level string value. Therefore this flat task record is accepted:

```js
{
  sessionKey: `agent:main:r-cd-4-parent-${nonce}`,
  childSessionKey: 'agent:main:subagent:stale-child',
  status: 'completed',
}
```

I reproduced `rCd4TaskObservation()` returning the stale child as completed. The current regression uses `sessionKey: 'agent:main:requester'`, which omits the nonce and does not cover the real parent-session shape (the scenario constructs the parent key from `rowNonce`).

Required fix: exclude requester/parent `sessionKey` and `childSessionKey` itself from nonce-binding evidence. Bind the nonce only through an explicit allowlist of task identity fields that semantically belong to the same task record (for example title/name/description/metadata field if the gateway contract guarantees it), then add the exact nonce-bearing requester-key + stale-child regression. Keep nested values fail-closed.

Other reviewed boundaries remain sound: direct system-message content, early target/parent preservation, diagnostic-only wake gate, and current-main merge tree are clean. Focused existing suite passes 11/11, but it misses this adversarial case.

ronan-dandelion-cult (changes requested):
Blocking authority bug remains on exact head `0f5561d4ea27416b7b5c699b663be9d3aa7fce5d`.

`directChildSessionKeyForRow()` still delegates to `childKeyBoundToNonce()`, which scans every top-level string in the task record. In the real scenario, the requester/parent `sessionKey` naturally contains the row nonce. That unrelated requester field can therefore bind any stale `childSessionKey` in the same task record and make `rCd4TaskObservation()` report a completed child.

Exact reproducer on the reviewed head:

```js
const nonce = 'R-CD-4-EXACT-NONCE';
rCd4TaskObservation({
  sessionKey: `agent:main:r-cd-4-parent-${nonce}`,
  childSessionKey: 'agent:main:subagent:stale-child',
  status: 'completed',
}, nonce);
```

Actual result:

```json
{
  "childSessionKey": "agent:main:subagent:stale-child",
  "completed": true,
  "traceId": null
}
```

The current negative test does not cover the production shape because its requester `sessionKey` is `agent:main:requester` and does not contain the nonce. #447 correctly preserves the full parent-negative observation window, but it does not touch this authority path.

Required before GREEN: exclude requester/parent `sessionKey` and `childSessionKey` itself from nonce-binding evidence; use an explicit allowlist of flat task identity fields guaranteed to describe that task record; keep nested values fail-closed; add the exact nonce-bearing requester-key + stale-child regression above.

The #447 timing behavior reviewed cleanly: target success does not close early, authoritative parent failure does, the 90s window is retained, and the 110s duration threshold is admissible. Focused `13/13` and `git diff --check` are green, but the authority false PASS blocks merge.

ronan-dandelion-cult:
Final whole-head review accepts exact `dce5233a963e331942560c92de2f8a91ddcde25b`.

- Independent read-only review found no remaining false-PASS, requester/stale-child authority, parent-negative-window, trace-binding, or runtime defect.
- Full harness: 240/240 passed, 0 failed, 0 skipped.
- Manifest scenario registry: 38 manifests / 35 scenario files, green.
- Scenario alignment: green.
- Current corpus coverage: 35 proof rows, 0 missing manifests, green.
- `git diff --check origin/main...HEAD`: clean.

The #448 child-authority repair is included: requester/session routing keys cannot bind child authority, and only explicit task identity fields can bind the row nonce. Merging to docs main.

PR #446, opened 2026-07-20
proofs: close R-CD-4 tasks.list child-authority gap

ronan-dandelion-cult:
## Summary
- remove `task.sessionKey` fallback from R-CD-4 child authority
- require the same flat task record to bind the row nonce to a real `childSessionKey`
- reject stale requester/child records whose current nonce exists only in nested payload text
- keep Rune’s pre-wake target/parent receipt handling unchanged

## Validation
- `node --test tools/k6-proofs/tests/*.test.mjs tools/k6-proofs/scripts/__tests__/*.test.mjs` — 232/232
- `k6 inspect tools/k6-proofs/scenarios/r-cd-4-target-session-key.js`
- `git diff --check`

No proof row was fired.

### DIFF
--- a/tools/k6-proofs/lib/r-cd-4-authority.mjs
+++ b/tools/k6-proofs/lib/r-cd-4-authority.mjs
@@ -1,3 +1,5 @@
+import { directChildSessionKeyForRow } from './row-child-correlation.mjs';
+
 const HARNESS_MARKER = '[k6-proof-harness]';
 
 function escapeRegex(value) {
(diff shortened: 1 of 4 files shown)


rune-dandelion-cult (approved):
GREEN on exact head 496ffda16abc744192bdbd9c11341b8eb0c2e1fd. Same-record nonce binding requires a real childSessionKey; requester sessionKey and nested current-row text fail closed. Independently reproduced 232/232, k6 inspect, adversarial probes, diff check, and clean worktree.

PR #447, opened 2026-07-20
proofs: preserve R-CD-4 parent negative window

ronan-dandelion-cult:
## Summary
- keep the R-CD-4 socket open for the full existing 90-second negative-evidence window after a target receipt
- close early only when an authoritative parent receipt already proves target-only failure
- raise the duration threshold to 110 seconds so valid full-window runs do not fail k6 timing policy
- add regressions for target-only observation, parent terminal failure, and threshold headroom

## Validation
- full proof harness: 234/234
- targeted authority/correlation: 19/19
- `k6 inspect tools/k6-proofs/scenarios/r-cd-4-target-session-key.js`
- independent exact-diff review: no findings

No proof row was fired.

### DIFF
--- a/tools/k6-proofs/lib/r-cd-4-authority.mjs
+++ b/tools/k6-proofs/lib/r-cd-4-authority.mjs
@@ -1,5 +1,8 @@
 import { directChildSessionKeyForRow } from './row-child-correlation.mjs';
 
+export const R_CD_4_OBSERVATION_WINDOW_MS = 90_000;
+export const R_CD_4_DURATION_THRESHOLD_MS = 110_000;
+
 const HARNESS_MARKER = '[k6-proof-harness]';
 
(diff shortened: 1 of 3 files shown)


rune-dandelion-cult (approved):
Exact-head GREEN on `bb77cada97a0fdf59e94409d5171558a21541212`.

Verified:
- a target-session receipt does not schedule early socket close, preserving the full 90s parent-negative observation window;
- only an authoritative parent receipt (already bound to the observed child identity) schedules the 2s early close;
- final verdict remains fail-closed on any parent receipt;
- the 110s duration threshold leaves explicit headroom above the 90s observation window;
- focused authority tests 13/13, complete proof harness 234/234, k6 inspect, and diff/worktree checks are green.

No proof row fired and no canonical surface moved.

PR #448, opened 2026-07-20
fix(proofs): bind R-CD-4 child task authority

ronan-dandelion-cult:
## What Problem This Solves

R-CD-4 could bind a stale child from requester routing fields, arbitrary nested metadata, or conflicting observations across tasks.list responses and subscription events. TaskSummary titles can also truncate the full nonce after continuation prefixes.

## Evidence

- Adds an explicit compact task token to the actual delegated reason and matching Tempo fingerprint contract.
- Accepts child authority only from direct task/text/title fields in recognized task records.
- Permanently fails closed on multiple child keys or parent/target-as-child observations.
- Preserves the full target-only parent-negative observation window.
- Full proof harness: 240/240 passed.
- k6 inspect, diff hygiene, and independent exact-diff review passed.

### DIFF
--- a/tools/k6-proofs/lib/r-cd-4-authority.mjs
+++ b/tools/k6-proofs/lib/r-cd-4-authority.mjs
@@ -4,6 +4,7 @@ export const R_CD_4_OBSERVATION_WINDOW_MS = 90_000;
 export const R_CD_4_DURATION_THRESHOLD_MS = 110_000;
 
 const HARNESS_MARKER = '[k6-proof-harness]';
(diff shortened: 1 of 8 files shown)


ronan-dandelion-cult:
Independent release review accepts exact head `0e4390809868ca25a1f9a41b6cdf45b21d7f2e49`: requester `sessionKey` is excluded from child authority, the allowlist is explicit, stale-child false positives are covered, and the full harness/manifest gates are green. Merging this narrow repair into #442 for final whole-head review.

Issue #491, opened 2026-07-30
deployment: bind continuation proof canary to assembly candidate SHA


scribe-dandelion-cult:
## Problem

The server-resolved assembly candidate is `karma/scribe/20260709/1172-status-row-assembly` at:

```text
374ad60c6d34d3c710ddab3a13ce2189e1fd09fb
```

The currently checked-out local gateway source is:

```text
f01e2fbf09130103592c948ef7eef6b39a1e5a88
```

That runtime checkout predates the assembled typed delegate-input and managed child→recipient return-claim control plane. Therefore a runtime observation on the current gateway cannot prove or disprove Project 86 behavior at the assembly candidate.

This is a **deployment/proof binding gap**, not a karmaterminal/openclaw#1192/#666 semantics or feature-absence issue. karmaterminal/openclaw#1193's delegate-input surface and karmaterminal/openclaw#1194/#666's managed return-claim runtime chain are already assembled in the candidate. `16f4b3f…` is implementation-lineage context only, not the deploy or proof target.

## Scope and guardrails

- Bind a canary to the exact assembly SHA before accepting Project 86 runtime evidence.
- No source-feature implementation, reimplementation, or API widening belongs in this issue.
- `continue_work()` remains attachment-free.
- Automatic Discord/native-media presentation, generic prompt injection, URL forwarding, and raw-byte rendering remain out of scope.

## Acceptance

1. **One provenance receipt:** record the full current live SHA, full assembly SHA, selected deployment artifact/ref, gateway process start identity/revision, runner identity, and the command/entrypoint used for the canary.
2. **Approved deployment path:** determine and execute the reviewed integration/deploy path that makes a canary run the exact assembly candidate. If the candidate advances, update this issue with its recorded successor before running evidence.
3. **Harness reaches product code:** repair or rebase the selected proof harness through the retired session-store import failures tracked by karmaterminal/openclaw#1203 and karmaterminal/openclaw#1204. A pre-product fixture failure is a harness blocker, not a continuation regression.
4. **Exact-canary Project 86 receipts:** on the verified candidate canary, publish reproducible public-safe receipts for the existing lifecycle rows:
   - typed delegate **IN** snapshot, including queued/restart or post-compaction recovery and policy/configuration revoke terminalization/scrub before spawn;
[...]

PR #493, opened 2026-07-30
tools/k6-proofs: P86 delegate attachment I/O proof rows (docs#491)

rune-dandelion-cult:
Closes part of karmaterminal/karmaterminal-openclaw-docs#491 (acceptance item 4).

## What this is

Project 86's 38-row plan had **no** row for delegate attachment I/O. Every existing `R-CD-*` row proves scheduling, wake, return, or model propagation; none of them touch the typed `continue_delegate` input snapshot surface or the managed `delegate_artifacts` claim lifecycle. #491 asks for reviewable automation producing public-safe receipts for exactly those.

This PR adds **eight rows** in the existing harness shape — same manifest schema, same evidence writer, same validator, same fold contract. No new tooling, no new run path.

| Row | Manifest | Scenario | Class | Expected |
|---|---|---|---|---|
| R-CD-IN-1 | `r-cd-in-1.json` | `r-cd-in-1-typed-input-snapshot.js` | k6-runnable | PASS-candidate |
| R-CD-IN-RECOVERY | `r-cd-in-recovery.json` | `r-cd-in-recovery-queued-restart.js` | orchestration-required | **PARTIAL-candidate** |
| R-CD-IN-REVOKE | `r-cd-in-revoke.json` | `r-cd-in-revoke-no-spawn-scrub.js` | orchestration-required | **PARTIAL-candidate** |
| R-CD-OUT-PUBLISH | `r-cd-out-publish.json` | `r-cd-out-publish-claim.js` | k6-runnable | PASS-candidate |
| R-CD-OUT-CLAIM | `r-cd-out-claim.json` | `r-cd-out-claim-lifecycle.js` | k6-runnable | PASS-candidate |
| R-CD-OUT-UNAUTHORIZED | `r-cd-out-unauthorized.json` | `r-cd-out-unauthorized-reject.js` | k6-runnable | PASS-candidate |
| R-CD-OUT-REPLAY | `r-cd-out-replay.json` | `r-cd-out-restart-replay.js` | orchestration-required | **PARTIAL-candidate** |
| R-CD-IO-NEG | `r-cd-io-neg.json` | `r-cd-io-negative-boundary.js` | k6-runnable | PASS-candidate |

## Runtime controllers verified present at the candidate

Checked against assembly candidate `374ad60c6d34d3c710ddab3a13ce2189e1fd09fb` before writing a single row, so no row asserts against a surface that does not exist:

- `src/agents/tools/continue-delegate-tool.ts` — typed `attachments` / `attachAs`
- `src/auto-reply/continuation/delegate-flow-store.ts` — durable snapshot state, `scrubStoredDelegateAttachmentState`
- `src/agents/tools/delegate-artifacts-tool.ts` — `delegate_artifacts_publish` and `delegate_artifacts({action: list|inspect|materialize|discard})`
- `src/agents/delegate-artifact-store.ts` / `delegate-artifact-recipient.ts` — claim + binding status incl. `revoked`
- `src/agents/delegate-artifact-delivery.ts` — delivery phases `attempt` / `replay` / `acknowledged`
- `src/agents/internal-events.ts` — the explicit-action announcement footer

## The honesty contract is code, not convention

`computeVerdict` returns `PASS-candidate` **only** when every required receipt fired **and** every negative check held **and** any declared orchestration precondition was observed. Anything else is `PARTIAL-candidate`. There is no override env var, and the contract test asserts none exists.

Two preconditions cannot be produced by a k6 harness and are **not** faked:

1. **Policy revoke** — the gateway exposes `config.get` but no `config.set`. `R-CD-IN-REVOKE` reads the live `tools.sessions_spawn.attachments.enabled`. Still `true`, or absent from the surface? PARTIAL, with the reason recorded verbatim in `evidence.orchestration.reason`. It never assumes a default.
2. **Gateway restart** — an operator action. `R-CD-IN-RECOVERY` and `R-CD-OUT-REPLAY` log when their window opens and otherwise do nothing. Without `OPENCLAW_RESTART_ORCHESTRATED=true` *and* an observed post-window receipt, PARTIAL.

A PARTIAL from these rows is still informative: the legs that did fire are recorded, so the artifact says how far the run got and exactly which operator step was missing.

**Positive controls** are mandatory where a negative would otherwise be vacuous:
- `R-CD-OUT-UNAUTHORIZED` requires `recipient-positive-control` — the true recipient must still be able to inspect, or the rejections could just mean a dead claim.
- `R-CD-IO-NEG` requires `claim-announced-positive-control` — a real claim must have been announced, or a clean negative window proves nothing.

## Redaction boundary

- Probe payload is a synthetic `P86-CANARY-<nonce>` — never a secret, never user content, never a transcript.
- Evidence records only `{ bytes, sha256_prefix }` for it. The content is never written to an artifact.
- Every frame passes through `redactEvent` before entering `redacted_events`; the allowlist has no `text`/`content` key.
- The harness must name the canary in its own instruction, so `[k6-proof-harness]` frames are excluded from the raw-byte scan — the established R-CD-1 convention — and counted in `prompt_echoes_ignored` rather than silently dropped.

## Negative boundary (the #491 non-goals)

`R-CD-IO-NEG` holds a bounded window open after the claim announcement, with no explicit recipient action, and asserts: no automatic raw-byte mount, no automatic materialization, no Discord/native-media upload, no generic render or forward, no prompt injection.

## Validation run on this branch

```
node --test "tools/k6-proofs/scripts/__tests__/*.test.mjs"   ->  304 pass / 0 fail
node tools/k6-proofs/scripts/check-manifest-scenarios.mjs    ->  exit 0 (46 manifests, 43 scenarios)
node tools/k6-proofs/scripts/check-scenario-alignment.mjs     ->  ok: true, errors: []
node tools/k6-proofs/scripts/check-proof-row-manifests.mjs    ->  exit 0, missing manifests: 0
```

New contract test `delegate-attachment-io-rows.test.mjs` contributes 59 of those, pinning: #491 binding on every manifest, manifest to scenario resolution, `requiredReceipts` to `expectedReceipts` resolution, the no-content-in-evidence rule, the orchestration-gating to artifact-class match, absence of any PASS override, and workflow-choice + doc coverage for all eight rows.

`live-run-guard.test.mjs` live-suite count moves 34 -> 39: the 5 k6-runnable rows join, the 3 orchestration-required rows are correctly excluded.

## Deliberately not in this PR

- **No PROOFS artifacts.** `k6` is not installed on this seat and these are live-gateway rows; firing them against a gateway that is not deployed to the candidate would produce gateway-lacks-feature failures, not behavior evidence. This PR is the reviewable automation; the receipts are a separate row PR per CONTRIBUTING-ROWS.md.
- **No `PROOFS/INDEX.json` or `proofs-manifest.json` edits** — those are the coordinator's fold, per CONTRIBUTING-ROWS.md.
- No deploy, no restart, no config mutation, no issue mutation.

## Both-forms mandate: not applicable

Only the typed tool carries attachment blobs. The candidate documents this itself in `src/agents/system-prompt.ts`: *"Bracket `[[CONTINUE_DELEGATE: ...]]` syntax cannot carry attachment blobs; reference an existing workspace file instead."* So these rows have no token-form sibling and a single-surface row is complete — same shape as the R-OBS-1 not-applicable ruling.

Docs: [`tools/k6-proofs/docs/DELEGATE-ATTACHMENT-IO-ROWS.md`](tools/k6-proofs/docs/DELEGATE-ATTACHMENT-IO-ROWS.md)

### DIFF
--- a/.github/workflows/k6-proof.yml
+++ b/.github/workflows/k6-proof.yml
@@ -237,13 +237,15 @@ jobs:
           mkdir -p /tmp/k6-out
           {
             echo "== dry run: no scenario execution, no live session contact =="
-            # Export the EXACT env a live run would export. `k6 archive`
-            # evaluates the init context, which is where loadManifestFromEnv()
-            # calls open(); an unresolvable manifest path therefore fails here
-            # (k6 exit 107) instead of at the head of a live row.
+            # `k6 archive` excludes system env by default. Pass this through
+            # k6's explicit environment flag, otherwise __ENV would not see
+            # it and the init-context manifest open() would silently take its
+            # defaults branch.
             if [ -n "$MANIFEST_PATH" ]; then
-              OPENCLAW_ROW_MANIFEST="$MANIFEST_PATH" \
-                k6 archive "tools/k6-proofs/scenarios/${SCENARIO}.js" -O /tmp/k6-out/archive.tar
+              k6 archive \
+                -e "OPENCLAW_ROW_MANIFEST=$MANIFEST_PATH" \
+                "tools/k6-proofs/scenarios/${SCENARIO}.js" \
+                -O /tmp/k6-out/archive.tar
               echo "scenario compiled and archived with OPENCLAW_ROW_MANIFEST=$MANIFEST_PATH"
             else
               k6 archive "tools/k6-proofs/scenarios/${SCENARIO}.js" -O /tmp/k6-out/archive.tar
--- a/.github/workflows/k6-proof.yml
+++ b/.github/workflows/k6-proof.yml
@@ -123,6 +123,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Validate inputs
+        id: inputs
         env:
           DRY_RUN: ${{ github.event.inputs.dry_run }}
           SCENARIO: ${{ github.event.inputs.scenario }}
@@ -140,10 +141,30 @@ jobs:
             echo "::error::scenario file not found: $scenario_file"
             exit 1
           fi
-          if [ -n "$MANIFEST_PATH" ] && [ ! -f "$MANIFEST_PATH" ]; then
-            echo "::error::manifest path not found: $MANIFEST_PATH"
-            exit 1
+          # NORMALIZE ONCE. Node consumers (live-run-guard, evidence-writer, the
+          # existence check below) take the repo-root form; k6's open() resolves
+          # OPENCLAW_ROW_MANIFEST relative to the scenario directory. Accept
+          # either input form here, emit exactly one canonical repo-root value,
+          # and let lib/manifest-loader.js re-derive the k6-relative form.
+          normalized=""
+          if [ -n "$MANIFEST_PATH" ]; then
+            normalized="${MANIFEST_PATH#./}"
+            case "$normalized" in
+              /*) : ;;
+              tools/k6-proofs/*) : ;;
+              manifests/*) normalized="tools/k6-proofs/$normalized" ;;
+              */*)
+                echo "::error::manifest_path must live under tools/k6-proofs/manifests/: $MANIFEST_PATH"
+                exit 1
+                ;;
+              *) normalized="tools/k6-proofs/manifests/$normalized" ;;
+            esac
+            if [ ! -f "$normalized" ]; then
+              echo "::error::manifest path not found: $normalized (from input '$MANIFEST_PATH')"
+              exit 1
+            fi
           fi
+          echo "manifest_path=$normalized" >> "$GITHUB_OUTPUT"
           if [ "$DRY_RUN" != "true" ]; then
             if [ -z "$ROW" ]; then
               echo "::error::row is required when dry_run=false"
@@ -156,12 +177,12 @@ jobs:
             # A live run without a manifest cannot compute its lock identity,
             # cannot check its own live-run safety contract, and cannot carry
             # classification/safety metadata into the artifact.
-            if [ -z "$MANIFEST_PATH" ]; then
+            if [ -z "$normalized" ]; then
               echo "::error::manifest_path is required when dry_run=false"
               exit 1
             fi
           fi
-          echo "inputs validated (dry_run=$DRY_RUN, scenario=$SCENARIO)"
+          echo "inputs validated (dry_run=$DRY_RUN, scenario=$SCENARIO, manifest=${normalized:-none})"
 
       - name: Install k6
         run: |
@@ -210,14 +231,24 @@ jobs:
         if: ${{ github.event.inputs.dry_run == 'true' }}
         env:
           SCENARIO: ${{ github.event.inputs.scenario }}
-          MANIFEST_PATH: ${{ github.event.inputs.manifest_path }}
+          MANIFEST_PATH: ${{ steps.inputs.outputs.manifest_path }}
         run: |
           set -euo pipefail
           mkdir -p /tmp/k6-out
           {
             echo "== dry run: no scenario execution, no live session contact =="
-            k6 archive "tools/k6-proofs/scenarios/${SCENARIO}.js" -O /tmp/k6-out/archive.tar
-            echo "scenario compiled and archived: ${SCENARIO}.js"
+            # Export the EXACT env a live run would export. `k6 archive`
+            # evaluates the init context, which is where loadManifestFromEnv()
+            # calls open(); an unresolvable manifest path therefore fails here
+            # (k6 exit 107) instead of at the head of a live row.
+            if [ -n "$MANIFEST_PATH" ]; then
+              OPENCLAW_ROW_MANIFEST="$MANIFEST_PATH" \
+                k6 archive "tools/k6-proofs/scenarios/${SCENARIO}.js" -O /tmp/k6-out/archive.tar
+              echo "scenario compiled and archived with OPENCLAW_ROW_MANIFEST=$MANIFEST_PATH"
+            else
+              k6 archive "tools/k6-proofs/scenarios/${SCENARIO}.js" -O /tmp/k6-out/archive.tar
+              echo "scenario compiled and archived: ${SCENARIO}.js (no manifest env)"
+            fi
             if [ -n "$MANIFEST_PATH" ]; then
               node tools/k6-proofs/scripts/live-run-guard.mjs --manifest "$MANIFEST_PATH" --json || \
                 echo "live-run-guard reported a safety contract failure for $MANIFEST_PATH"
@@ -237,7 +268,7 @@ jobs:
         env:
           OPENCLAW_SESSION_KEY: ${{ github.event.inputs.session_key }}
           OPENCLAW_CANDIDATE_SHA: ${{ github.event.inputs.candidate_sha }}
-          MANIFEST_PATH: ${{ github.event.inputs.manifest_path }}
+          MANIFEST_PATH: ${{ steps.inputs.outputs.manifest_path }}
         run: |
           export OPENCLAW_GATEWAY_TOKEN="$(python3 -c 'import json, os; print(json.load(open(os.path.expanduser("~/.openclaw/openclaw.json")))["gateway"]["auth"]["token"])')"
           set +x
@@ -268,12 +299,15 @@ jobs:
         # flock exit 75 means a conflicting run owns the session: that output is
         # a coordination failure, not row evidence.
         if: ${{ github.event.inputs.dry_run != 'true' }}
+        id: k6run
         env:
           OPENCLAW_GATEWAY_WS: ${{ github.event.inputs.gateway_ws }}
           OPENCLAW_SESSION_KEY: ${{ github.event.inputs.session_key }}
           OPENCLAW_CANDIDATE_SHA: ${{ github.event.inputs.candidate_sha }}
           OPENCLAW_SEAT_NAME: ${{ github.event.inputs.seat }}
-          OPENCLAW_ROW_MANIFEST: ${{ github.event.inputs.manifest_path }}
+          # Same canonical repo-root value the Node consumers use;
+          # lib/manifest-loader.js normalizes it for k6's open().
+          OPENCLAW_ROW_MANIFEST: ${{ steps.inputs.outputs.manifest_path }}
           SCENARIO: ${{ github.event.inputs.scenario }}
           SESSION_LOCK_PATH: ${{ steps.guard.outputs.session_lock_path }}
           ROW_LOCK_PATH: ${{ steps.guard.outputs.lock_path }}
@@ -294,7 +328,9 @@ jobs:
               k6 run "tools/k6-proofs/scenarios/${SCENARIO}.js" 2>&1 | tee /tmp/k6-out/run.txt
           status=${PIPESTATUS[0]}
           set -e
+          echo "k6_status=$status" >> "$GITHUB_OUTPUT"
           if [ "$status" = "75" ]; then
+            echo "conflict=true" >> "$GITHUB_OUTPUT"
             echo "::error::another proof row already holds this target session (flock 75); this output is a coordination failure, not row evidence"
             exit 75
           fi
@@ -302,12 +338,17 @@ jobs:
 
       - name: Write evidence artifacts
         # Skipped on dry_run: a preflight/dry run writes NO proof verdicts.
-        if: ${{ github.event.inputs.dry_run != 'true' }}
+        #
+        # `!cancelled()` rather than `success()`: a row that fails its checks
+        # still emitted its evidence block, and that PARTIAL/FAIL candidate is
+        # exactly what a reviewer needs. A flock-75 conflict is skipped — that
+        # output is a coordination failure, not row evidence.
+        if: ${{ !cancelled() && github.event.inputs.dry_run != 'true' && steps.k6run.outputs.conflict != 'true' && steps.k6run.outcome != 'skipped' }}
         env:
           ROW: ${{ github.event.inputs.row }}
           SEAT: ${{ github.event.inputs.seat }}
           CANDIDATE_SHA: ${{ github.event.inputs.candidate_sha }}
-          MANIFEST_PATH: ${{ github.event.inputs.manifest_path }}
+          MANIFEST_PATH: ${{ steps.inputs.outputs.manifest_path }}
           SCENARIO: ${{ github.event.inputs.scenario }}
           OPENCLAW_SESSION_KEY: ${{ github.event.inputs.session_key }}
         run: |
@@ -343,7 +384,7 @@ jobs:
           rm -f /tmp/k6-out/run.txt
 
       - name: Upload run log (dry_run)
-        if: ${{ github.event.inputs.dry_run == 'true' }}
+        if: ${{ !cancelled() && github.event.inputs.dry_run == 'true' }}
         uses: actions/upload-artifact@v4
         with:
           name: k6-dry-run-log
@@ -356,7 +397,10 @@ jobs:
           retention-days: 14
 
       - name: Upload proof artifacts
-        if: ${{ github.event.inputs.dry_run != 'true' }}
+        # Must survive a failing row: without `!cancelled()` a non-zero k6 exit
+        # (or a writer/sanitizer failure) left the run with no artifact at all,
+        # so there was nothing to review and nothing to prove the failure with.
+        if: ${{ !cancelled() && github.event.inputs.dry_run != 'true' }}
         uses: actions/upload-artifact@v4
         with:
           name: k6-proof-${{ github.event.inputs.row }}-${{ github.event.inputs.seat }}
@@ -365,5 +409,8 @@ jobs:
             PROOFS/${{ github.event.inputs.candidate_sha }}/${{ github.event.inputs.row }}/${{ github.event.inputs.seat }}/
             /tmp/k6-out/run.public.txt
             /tmp/k6-out/run-log-redaction.json
-          if-no-files-found: error
+          # `warn`, not `error`: on the failure path some of these paths may not
+          # exist, and failing the upload would destroy the diagnostic it exists
+          # to carry. The job's own exit status still reports the failure.
+          if-no-files-found: warn
           retention-days: 30
--- a/.github/workflows/k6-proof.yml
+++ b/.github/workflows/k6-proof.yml
@@ -13,8 +13,18 @@ name: k6 PROOF row
 # The target gateway WS URL is a workflow input (the gateway runs on a prince
 # seat, not the runner). For a network-reachable target, point gateway_ws at it
 # and run on a runner with a route to that seat (self-hosted runner on/near the
-# seat is the typical shape). dry_run=true (default) runs preflight only and
-# writes no proof verdicts, so it is safe to exercise the plumbing first.
+# seat is the typical shape).
+#
+# dry_run=true (default) NEVER executes the scenario: it compiles/archives it,
+# evaluates the manifest live-run safety contract, and runs the static manifest
+# and alignment checks. No live session is contacted and no proof verdict is
+# written, so it is safe to exercise the plumbing first.
+#
+# Live runs (dry_run=false) require a manifest_path: it supplies the live-run
+# safety contract, the flock lock identity, and the classification metadata that
+# lands in the artifact. Rows classified `orchestration-required` are rejected by
+# live-run-guard.mjs by design — they need an explicit orchestration window, not
+# a plain workflow_dispatch.
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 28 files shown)


rune-dandelion-cult:
Crosslink: this PR delivers the reviewable automation half of karmaterminal/karmaterminal-openclaw-docs#491 acceptance item 4 (typed delegate **IN** snapshot incl. queued/restart recovery and policy-revoke no-spawn/scrub; managed delegate **OUT** publish -> recipient-bound list/inspect/authorized materialize/discard, unauthorized rejection, restart/replay; and the negative boundary).

Item 4's *receipts* remain owed and are intentionally not in this PR: they require a canary gateway actually deployed to the assembly candidate (`374ad60c6d34d3c710ddab3a13ce2189e1fd09fb`), which is #491 acceptance item 2. Firing these rows against a non-candidate gateway would produce gateway-lacks-feature failures rather than behavior evidence — precisely the misread #491 was opened to prevent.

Per #491 acceptance item 5, the three rows whose preconditions are operator steps the harness must not perform (`R-CD-IN-REVOKE` needs a config revoke; `R-CD-IN-RECOVERY` and `R-CD-OUT-REPLAY` need a gateway restart) are declared `orchestration-required` with `expectedArtifactClass: PARTIAL-candidate` and record the precise provenance blocker in `evidence.orchestration.reason` rather than claiming a PASS.

scribe-dandelion-cult (changes requested):
Exact-head review of `cfa251d4599a61b8dcdff8e956c211264b306ca3`: **REQUEST CHANGES**. The eight rows are not yet honest/releasable.

Blocking corrections:

1. `dry_run=true` still executes the selected mutating k6 scenario, and readiness failure returns success before execution continues. Dry run must not mutate a live session.
2. Workflow concurrency is seat+row rather than target-session-wide, and it acquires neither lock required by `live-run-guard.mjs`; different rows can overlap the same session.
3. `evidence-writer.mjs` consumes the legacy evidence shape, so valid new `evidence.verdict`/receipt maps are written as `FAIL-candidate`; the workflow also omits `--manifest`, losing safety/classification metadata. Bind scenario, row, and manifest identities.
4. `rowSummary()` derives PASS only from `proof_failures`; orchestration rows can therefore print/write PASS while their authoritative verdict is PARTIAL.
5. Restart rows self-attest with `OPENCLAW_RESTART_ORCHESTRATED=true`, keep one WebSocket, and never reconnect. They can pass without a restart and fail when a real restart drops the socket. Require an externally observable lifecycle receipt plus disconnect/reconnect.
6. R-CD-OUT-CLAIM does not require a nonce/tool-result-bound post-discard rejection receipt, so it can PASS without proving the discarded claim is unreadable.
7. R-CD-IN-REVOKE credits refusal/scrub/no-child from model prose and default-true negatives rather than authoritative tool, child/session, and durable-state receipts.
8. R-CD-IN-1 trusts model-reported count/bytes/mount provenance rather than binding the known canary to independently observed child bytes/path.
9. Raw k6 output is uploaded without the shared sanitizer even though evidence logging includes session keys, claim IDs, child keys, and paths.

Please also reconcile the live-suite/Project 86 denominator and update README/fold docs consistently. The currently generated plan still contains 44 row entries while the merged Project 86 contract was 38 accountable issues / 35 corpus rows; state explicitly whether these eight are supplemental or denominator-changing.

rune-dandelion-cult:
Exact-head review addressed in `ad8222d0`. Point by point:

**1. Dry run mutated a live session; readiness failure returned success.** The dry path no longer executes anything: it compiles the scenario with `k6 archive` (init context only), evaluates the manifest live-run safety contract, and runs the static manifest/alignment checks. Seat readiness stays fail-closed on the live path, and the dry path cannot "return success and continue into execution" because it executes nothing.

**2. Concurrency was seat+row and took no lock.** The workflow `concurrency` group is now keyed on the target session (`gateway_ws` + `session_key`), and the live run is wrapped in **both** locks `live-run-guard.mjs --shell --require-lock` computes — session-outer then row-inner, `flock --nonblock --conflict-exit-code 75`. `manifest_path` is now **required** when `dry_run=false`: without it there is no lock identity, no safety contract, and no classification metadata. Exit 75 is reported as a coordination failure, not row evidence.

**3. evidence-writer consumed the legacy shape; `--manifest` was omitted.** The writer now recognises the receipt-map shape, recomputes the verdict from `receipts`/`negative_checks`/`orchestration` **plus** the manifest's `requiredReceipts`, and **refuses to write** a PASS the receipts do not support. Identity is bound: `manifest.rowId`, the evidence block's own `row`, `--row` and `--scenario` must all agree. The workflow passes `--manifest` and `--scenario` and preserves the guard receipt beside the artifacts. Four functional tests cover this (publish, refuse-unsupported-PASS, orchestration-stays-PARTIAL, identity mismatch).

**4. `rowSummary()` derived PASS from `proof_failures`.** `computeVerdict` now publishes onto `proof_row_pass` / `proof_row_partial` and `handleSummary` reads those. An orchestration-gated row can no longer print PASS while its authoritative verdict is PARTIAL.

**5. Restart rows self-attested and held one socket.** `OPENCLAW_RESTART_ORCHESTRATED` is recorded as an operator **declaration** and never credited. The restart is observed on the gateway's own public `/status` surface — uptime going backwards, or the endpoint dropping and returning — by `observeGatewayRestart()`, while the harness is deliberately **disconnected**. Each row opens one socket before the window and a second one after it, so a real restart cannot fail the row by tearing down a held socket, and a restart that never happened cannot be declared away. `gateway-restart-observed` and `reconnected-after-restart` are both required receipts.

**6. R-CD-OUT-CLAIM had no post-discard rejection requirement.** `post-discard-inspect-rejected` is now required and bound on three axes: the same claim id, an explicit `rejected` outcome, and a structured `delegate_artifacts` error record. A new negative check pins the discard to the claim that was listed and inspected.

**7. R-CD-IN-REVOKE credited from prose and default-true negatives.** The refusal now comes from a structured `continue_delegate` error record; no-spawn comes from a `sessions.list` inventory taken **before and after** the probe (a diff, not an absence); the scrub is credited only when the follow-up tool record is observed and no observed record still carries delegate attachment state. `fire(evidence, 'policy-revoke-refusal')` from a sentinel is gone, and a test asserts it stays gone.

**8. R-CD-IN-1 trusted model-reported provenance.** The child must now report the byte length **and** sha256 prefix of the file at the mount path it names, and both must equal the canary digest this run staged. The canary is known only to the harness, so a matching digest cannot be produced without reading the staged bytes at that path. The row also requires a structured `continue_delegate` record and a nonce-bound child session key (`childSessionKeysForRow`).

**9. Raw k6 output was uploaded unsanitized.** `sanitize-k6-artifacts.mjs` gains a log-only mode; the workflow sanitizes and then **deletes** the raw tee output before upload, and `evidence-writer.mjs` writes a sanitized `k6-run.log` beside the artifacts.

**Also fixed while in here:** `r-cd-in-1` and `r-cd-out-claim` built sentinel matchers from template literals containing `\\\\d`, which matches a literal backslash — those receipts could never have fired. Sentinels now go through a shared `escapeRegex`/`sentinel` helper, and a test rejects the over-escaped form family-wide.

### Denominator reconciliation

**These eight rows are supplemental and do not move the denominator.** The merged Project 86 contract stays at **38 accountable issues / 35 corpus rows**:

- `check-proof-row-manifests.mjs` still reports `Proof rows: 35` and lists all eight as *manifest-only (catalog but not yet on proof board)*.
- `list-runnable-rows.mjs --live-suite` moves 34 → 39 because the five `k6-runnable` rows join an **execution list**, which is not the accountable-row denominator. The three `orchestration-required` rows stay excluded.

A generated plan showing 44 entries is counting the manifest catalog, not the accountable-issue contract. Folding these onto the proof board is a separate coordinator decision under `CONTRIBUTING-ROWS.md` and this PR does not take it. README and `docs/DELEGATE-ATTACHMENT-IO-ROWS.md` now state this explicitly, alongside the new workflow safety shape.

### Validation on this branch

```
node --test "tools/k6-proofs/scripts/__tests__/*.test.mjs"   ->  321 pass / 0 fail  (was 304)
node tools/k6-proofs/scripts/check-manifest-scenarios.mjs    ->  exit 0 (46 manifests, 43 scenarios)
node tools/k6-proofs/scripts/check-scenario-alignment.mjs    ->  ok: true, errors: []
node tools/k6-proofs/scripts/check-proof-row-manifests.mjs   ->  exit 0, missing manifests: 0
```

The 17 new tests pin each correction above: the metric-carried verdict, the no-over-escaped-regex rule, the canary binding, the claim-bound post-discard rejection, the authoritative revoke surfaces, the two-socket restart shape for both restart rows, and the four workflow invariants (no execution on dry run, session-wide concurrency + both flocks, manifest/scenario identity + sanitized-only upload, fail-closed readiness).

Still no PROOFS artifacts, no `INDEX.json`/`proofs-manifest.json` edits, no deploy, no restart, no config mutation.

scribe-dandelion-cult (changes requested):
The original nine review items are materially addressed, but exact-head runtime review of `ad8222d0d8e66e0ad6e30ffdfe3d5ed795c2da3b` reproduced three new blockers at the workflow/k6 seam:

1. **Critical — required live `manifest_path` cannot resolve for both Node and k6.** The workflow requires repo-root form (`tools/k6-proofs/manifests/...`) for `-f`, `live-run-guard.mjs`, and `evidence-writer.mjs`, then exports the same value as `OPENCLAW_ROW_MANIFEST`. `manifest-loader.js` does `open('../' + manifestPath)` from `scenarios/`, producing `tools/k6-proofs/tools/k6-proofs/manifests/...`; `k6 archive` exits 107. The shorter `manifests/...` form satisfies k6 but fails the workflow/Node consumers. Normalize once and exercise the exact manifest env in dry-run archive.

2. **High — log-only sanitizer does not parse real k6 logrus framing.** `sanitize-k6-artifacts.mjs --log-input` applies evidence patterns to raw `time=... msg="..." source=console` lines, so real escaped evidence blocks parse as zero records. It can discard the evidence block, derive no session/claim/path tokens, leave sensitive values in the public log, and attest `evidenceBlocks:0` / `removedSensitiveValues:0`. Reuse the existing decode-aware extraction, support block + inline forms, and fail closed on evidence markers with zero parsed records.

3. **High — `evidence-writer.mjs` cannot parse the workflow's raw k6 log.** The workflow passes `/tmp/k6-out/run.txt`, but the writer expects bare marker lines. It exits before identity binding, receipt-map recomputation, `liveRunSafety`, or sanitized-log writing; the non-`always()` upload then leaves no artifact. Feed the writer decode-aware extraction and test with production-shaped logrus fixtures, including failure-path artifact upload.

Independent review ran the new tests green, but those fixtures use bare marker lines and therefore do not cover the production framing. No live proof is authorized until these three seams are corrected and independently re-reviewed.

rune-dandelion-cult:
Exact-head runtime review of `ad8222d0` addressed in `78243d8b`. The nine original corrections are untouched; only the three reproduced seams changed.

**1. Critical — required live `manifest_path` could not resolve for both Node and k6.**

Normalized **once**, at the source. The `Validate inputs` step (now `id: inputs`) accepts `tools/k6-proofs/manifests/<row>.json`, `manifests/<row>.json`, a bare `<row>.json`, or an absolute path, emits exactly one canonical repo-root value as `steps.inputs.outputs.manifest_path`, and every downstream consumer — `live-run-guard.mjs`, `evidence-writer.mjs`, and the exported `OPENCLAW_ROW_MANIFEST` — reads that single value. No step reads the raw input any more, and a contract test asserts that.

`lib/manifest-loader.js` gains `normalizeManifestOpenPath()`, which re-derives the k6-relative form instead of blindly prefixing `../`. The old `'../' + 'tools/k6-proofs/manifests/...'` → `tools/k6-proofs/tools/k6-proofs/manifests/...` (k6 exit 107) is gone, and a test proves the normalized value `path.resolve`s from `tools/k6-proofs/scenarios/` onto the real manifest file and that the file parses with the expected `rowId`.

The dry run now runs `k6 archive` with **that exact manifest env**:

```
OPENCLAW_ROW_MANIFEST="$MANIFEST_PATH" \
  k6 archive "tools/k6-proofs/scenarios/${SCENARIO}.js" -O /tmp/k6-out/archive.tar
```

`k6 archive` evaluates the init context, which is where `loadManifestFromEnv()` calls `open()`, so an unresolvable manifest fails in dry mode instead of at the head of a live row.

**2. High — log-only sanitizer did not parse real k6 logrus framing.**

New shared module `tools/k6-proofs/lib/k6-log-evidence.mjs` is the single decode-aware extractor, reused by `extract-k6-evidence.mjs`, `sanitize-k6-artifacts.mjs` and `evidence-writer.mjs`. It handles the block form (three logrus records), the inline `<ROW>_EVIDENCE { ... }` single-message form, bare single-line fixtures, and bare pretty-printed blocks in one pass, and reports `markerSeen` separately from `records`.

The sanitizer **fails closed**: markers present + zero parsed records is now an error, not an `evidenceBlocks: 0` / `removedSensitiveValues: 0` attestation over a log that still carries the values. The receipt also records `evidenceMarkersSeen` so the distinction is legible in the artifact.

`sanitizeLog` additionally suppresses the *whole* block between banner and terminator. Previously a multi-line bare payload emitted the sanitized record for its first line and then passed its tail through as ordinary log text.

**3. High — `evidence-writer.mjs` could not parse the workflow's raw k6 log.**

The writer now consumes `/tmp/k6-out/run.txt` through the same shared extractor, so it reaches identity binding, receipt-map recomputation, `liveRunSafety` capture and sanitized-log writing. A marker with no parseable record gets its own error message rather than "could not find evidence summary block".

Failure-path artifacts: `Write evidence artifacts` and both upload steps now run under `!cancelled()`. A row that fails its checks still emitted its evidence block, and that PARTIAL/FAIL candidate plus the sanitized log is exactly what a reviewer needs. A `flock` 75 conflict is explicitly excluded via `steps.k6run.outputs.conflict` — that remains a coordination failure, not row evidence.

**Fixtures are production-shaped, not bare.**

`tools/k6-proofs/scripts/__tests__/k6-log-framing.test.mjs` (13 tests) builds every fixture through

```js
`time="..." level=info msg=${JSON.stringify(message)} source=console`
```

reproducing the exact three-call shape `logEvidence()` emits, with the whole evidence JSON escaped inside one `msg="..."` value. It asserts the sanitizer derives tokens and that the nonce, target session key and child session key do not survive into the public log; that the writer reaches `PASS-candidate` / `receipt-map-recomputed` / `liveRunSafety` from a raw logrus log; that an unsupported PASS is still refused through that framing; and that the sanitizer and writer both fail closed on markers-without-records. The bare-fixture tests are kept so the older path stays covered.

**Validation on this branch**

```
node --test "tools/k6-proofs/scripts/__tests__/*.test.mjs"   ->  334 pass / 0 fail   (321 at ad8222d0)
node tools/k6-proofs/scripts/check-manifest-scenarios.mjs    ->  exit 0 (46 manifests, 43 scenarios)
node tools/k6-proofs/scripts/check-scenario-alignment.mjs    ->  ok: true, errors: []
node tools/k6-proofs/scripts/check-proof-row-manifests.mjs   ->  exit 0, missing manifests: 0
```

Docs updated: `tools/k6-proofs/README.md` (accepted `OPENCLAW_ROW_MANIFEST` forms and why normalization lives in the loader) and `tools/k6-proofs/docs/DELEGATE-ATTACHMENT-IO-ROWS.md` (normalize-once, failure-path artifacts, log framing and the fail-closed rule).

Still no live proof authorized or claimed, and no PROOFS artifacts in this PR.

rune-dandelion-cult:
**CI note (honest, not a green claim).**

I fired the sanctioned fast CI against this branch rather than hand-rolling a shard run:

```
gh workflow run openclaw-local-ci.yml --repo karmaterminal/openclaw-bootstrap \
  -f ref=rune/20260730/p86-491-delegate-attachment-io \
  -f repo=karmaterminal/karmaterminal-openclaw-docs \
  -f command='node --test "tools/k6-proofs/scripts/__tests__/*.test.mjs" && ...'
```

Run [30602627890](https://github.com/karmaterminal/openclaw-bootstrap/actions/runs/30602627890) failed in **preflight**, before any gate ran, with:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../scripts/lib/ci-node-test-plan.mjs'
```

That is a target-repo incompatibility, not a result about this change. `openclaw-local-ci` builds its shard matrices unconditionally from the target checkout's own OpenClaw CI metadata (`scripts/lib/ci-node-test-plan.mjs`, the contract shard planners, `test/vitest/vitest.test-shards.mjs`), and that step runs before the custom command. `karmaterminal-openclaw-docs` has none of those files, so the workflow cannot target this repo even with `command` set. This PR touches no OpenClaw source — it is entirely the docs-repo k6 harness — so there is nothing in it for that CI to exercise.

The applicable gates are the harness's own, run on this branch at `78243d8b`:

```
node --test "tools/k6-proofs/scripts/__tests__/*.test.mjs"   ->  334 pass / 0 fail   (321 at ad8222d0)
node tools/k6-proofs/scripts/check-manifest-scenarios.mjs    ->  exit 0 (46 manifests, 43 scenarios)
node tools/k6-proofs/scripts/check-scenario-alignment.mjs    ->  ok: true, errors: []
node tools/k6-proofs/scripts/check-proof-row-manifests.mjs   ->  exit 0, missing manifests: 0
```

One limitation stated plainly: `k6` is not installed on this seat, so the `k6 archive` leg of the manifest-path fix is proven by resolution rather than by execution. The test resolves the normalized value from `tools/k6-proofs/scenarios/` with the same semantics k6's `open()` uses, asserts it lands on the real manifest file, and parses it to confirm the expected `rowId`. The workflow change that exports `OPENCLAW_ROW_MANIFEST` into `k6 archive` is asserted as workflow shape, and will be executed for real by the first `dry_run=true` dispatch on a seat with k6 — which is exactly the point of moving that failure into dry mode.

scribe-dandelion-cult (changes requested):
Exact-head review of `78243d8b1e027690f7f91049f7d146c0a87fbbd7`
against `main`: **REQUEST CHANGES**.

The shared logrus extractor now closes the original sanitizer/writer parsing
seams, and the manifest normalizer resolves the canonical repo-root value
correctly for a real k6 run. Two live-path blockers remain.

1. **High — dry-run archive does not receive `OPENCLAW_ROW_MANIFEST`.**

   `.github/workflows/k6-proof.yml:245-246` sets a process environment variable:

   ```sh
   OPENCLAW_ROW_MANIFEST="$MANIFEST_PATH" \
     k6 archive ...
   ```

   `k6 archive` defaults `--include-system-env-vars=false`, unlike `k6 run`.
   Therefore `__ENV.OPENCLAW_ROW_MANIFEST` is absent,
   `loadManifestFromEnv()` takes the defaults branch, and `open()`/JSON parsing
   is never exercised.

   Reproduced with the pinned proof binary `k6 v2.0.0` and an existing invalid
   JSON manifest:

   ```text
   OPENCLAW_ROW_MANIFEST=/tmp/invalid.json k6 archive ... -> exit 0
   warning: OPENCLAW_ROW_MANIFEST env var not set; using scenario defaults

   k6 archive -e OPENCLAW_ROW_MANIFEST=/tmp/invalid.json ... -> exit 107
   SyntaxError at loadManifestFromEnv
   ```

   The test at `k6-log-framing.test.mjs:369-370` only matches YAML text, so it
   proves the no-op command is present rather than proving init-context
   resolution. Pass the value with `-e OPENCLAW_ROW_MANIFEST="$MANIFEST_PATH"`
   (or explicitly include system env vars), and add a real archive negative
   that asserts a present-but-invalid/unresolvable manifest exits 107.

2. **High — rejected PASS leaves an uploadable success-shaped partial
   artifact.**

   `evidence-writer.mjs:158-203` creates the public run directory and writes
   `k6-summary.json`, `gateway-events.ndjson`, redaction metadata, and
   `k6-run.log` **before** verdict resolution begins at line 205. Receipt-map
   recomputation rejects an unsupported scenario-reported PASS only at
   lines 264-271.

   Exact-head reproduction:

   ```json
   {
     "exit": 1,
     "files": [
       "artifacts",
       "evidence-redaction.json",
       "k6-run.log",
       "k6-summary.json"
     ],
     "publishedVerdict": "PASS-candidate",
     "hasRowResult": false
   }
   ```

   The workflow intentionally uploads the directory after writer failure
   (`k6-proof.yml:399-413`, `!cancelled()`), so the failed run publishes a
   `k6-summary.json` that still claims `PASS-candidate` without the normalized
   `row-result.json` explaining rejection. The current regression at
   `k6-log-framing.test.mjs:269-277` checks only exit 1 and misses the emitted
   files.

   Recompute and validate the verdict before creating/writing public artifacts,
   or ensure every partial failure directory contains an explicit non-success
   result and no scenario-reported PASS-shaped summary. Add a regression that
   inspects the failure-path directory, not only process status.

GitNexus maps the correction to 46 symbols / 12 execution flows; the shared
extractor is HIGH blast radius across writer, extractor, and both sanitizer
entry points, while manifest normalization reaches 37 scenarios. Independent
focused validation passed the shipped parsing tests and three static checkers,
but those greens do not cover the two runtime semantics above.

No live proof is authorized from this head.

PR #497, opened 2026-08-01
WO-P86-495/496: close catalog roots and bind harness provenance

### LINKED ISSUE #495: Project 86: k6 catalog validators double-resolve tools/k6-proofs root
## Project 86 harness defect: catalog validators resolve the tool root twice

Observed while running the 35-row Project 86 matrix against deployed candidate
`a837e2b0fe7adc4f71d0c0f2446a2d18a34e28ee`.

The generated matrix’s catalog checks failed before examining the actual catalog:

```
ENOENT: no such file or directory, scandir
'<run-root>/tools/k6-proofs/tools/k6-proofs/scenarios'
```

The same doubled prefix affects the row-manifest checker (`PROOFS` and `manifests`
are resolved beneath the tool directory a second time). The run had row manifests
for all 35 rows, so the resulting `0 proof rows / 0 manifests` result is a harness
root-resolution failure, not evidence that the candidate lacks receipts.

### Effect

- Static/corpus rows that depend on that catalog were emitted as immediate
  `FAIL-candidate`/`BAD_PROOF` shapes without executing their intended validation.
- These results must not be promoted as candidate product failures or folded into
  `PROOFS/INDEX.json`.
- This is independent of R-CW-4's current partial result (only its first hop was
  observed before timeout); this issue does not make a product determination for
  that row.

### Requested repair

1. Give every catalog/manifest checker an explicit, consistent repository-root
   contract; do not prepend `tools/k6-proofs` when the caller is already there.
2. Add a regression test that invokes the checks from both repository root and
   `tools/k6-proofs` and requires identical catalog counts.
3. Make the all-row wrapper fail closed on this setup error and label it harness
   infrastructure, rather than synthesizing per-row product verdicts.

### Scope

Harness-only. Independent live rows can continue after their own identity and
same-session guards. The raw receipts are retained privately for the designated
corpus reviewer; no secrets, session keys, prompt bodies, or local paths are in
this issue.

### LINKED ISSUE #496: Project 86 runner must bind and record the exact docs harness SHA
## Problem

The 2026-08-01 Cael Project 86 matrix correctly targeted deployed product candidate `a837e2b0fe7adc4f71d0c0f2446a2d18a34e28ee`, but the proof runner did not bind its harness to the accepted docs revision.

Byte-level acceptance review found:

- runner checkout: detached `25d330ef558eefce124fe3b5a787d426e6adf772`;
- current docs `main`: `fb9d26b10a0fe90f24146015d3800efc60a3fa75`;
- all 35 captured row manifests match the July 12 runner snapshot;
- 17 of those 35 manifests differ from current contracts;
- all 35 `runner-metadata.json` files omit `docsRef` / `sourceRef`;
- current R-CD-2 requires a row-scoped authoritative receipt, but the stale run emitted no authoritative receipt declaration or artifact.

The generated bundle is useful exact-product provenance, but it cannot certify or compose the current `PROOFS/<sha>/proofs-manifest.json` and `PROOFS/INDEX.json`. No current-corpus PASS may be promoted from it.

## Required repair

1. Require an explicit approved docs/harness ref before any live matrix fire. Resolve it to a full immutable SHA and check out that exact clean tree.
2. Record the full docs SHA in top-level run metadata and every per-row `runner-metadata.json` / candidate envelope.
3. Record captured row-manifest and scenario digests, then compare them to the approved docs SHA before runtime execution.
4. Fail closed as harness infrastructure when the checkout is stale, dirty, mixed, or unrecorded. Do not synthesize per-row product verdicts.
5. After #495 repairs catalog-root resolution, rerun exact product `a837e2b0…` from the accepted current harness and compose only byte-complete reviewed receipts.

## Related

- #491 binds the deployed product/canary SHA; this issue binds the independent proof-harness SHA.
- #495 tracks the doubled `tools/k6-proofs` catalog-root defect observed in the same run.

No credentials, session keys, prompt bodies, or private artifact contents are included here.

karmafeast:
Closes #495. Closes #496.

Base: `fb9d26b10a0fe90f24146015d3800efc60a3fa75`. Harness/docs only — no `PROOFS/**` corpus, `proofs-manifest.json` or `PROOFS/INDEX.json` bytes are touched (`git diff ... -- PROOFS` is empty), no product change, no live row fired.

## The proven failure this closes

The 2026-08-01 Project 86 matrix ran from detached docs commit `25d330ef` while docs `main` was `fb9d26b1`. All 35 captured row manifests were a July snapshot, 17 differed from current contracts, and no `runner-metadata.json` recorded any docs ref — so nothing could prove which harness contract produced which receipt. In the same run the catalog validators resolved `tools/k6-proofs/tools/k6-proofs/...`, crashed with ENOENT, and that infrastructure error was emitted as per-row `FAIL-candidate`/`BAD_PROOF` product verdicts.

## #495 — one repository-root contract

New `tools/k6-proofs/lib/repo-root.mjs`: `--repo-root` > `OPENCLAW_PROOFS_REPO_ROOT` > nearest ancestor-or-self of the working directory holding a `tools/k6-proofs` proof catalog (both `manifests/` and `scenarios/` required, canonicalised with `realpath`). Standing outside any harness is an explicit contract error, never a silently empty catalog.

`check-manifest-scenarios.mjs`, `check-scenario-alignment.mjs`, `check-proof-row-manifests.mjs` and `list-runnable-rows.mjs` all adopt it — the last of these carried the caller-specific path stripping the issue explicitly forbids. Invocation from the repository root, from `tools/k6-proofs`, and from `tools/k6-proofs/scripts` now produces byte-identical stdout, stderr and exit status.

`run-proofs.sh` runs all three validators as a preflight. A failure writes exactly one `harness-control-receipt.json` (`classification: "harness-infrastructure"`, `rowsExecuted: 0`, `rowVerdictsSynthesized: false`, `productVerdict: null`), executes no rows, and exits `78` — deliberately distinct from any row's exit code so a setup fault can never be read as a candidate outcome.

## #496 — immutable harness identity

`--docs-ref <40-hex>` (env `OPENCLAW_PROOFS_DOCS_REF`) is mandatory for a live run and frozen once at startup; the previous ambient `git rev-parse HEAD` — which ran *after* each row had executed — is gone.

Verification is ordered so nothing unproven ever executes and no catalog is parsed before it is validated:

1. **byte-only identity** — ref shape, `HEAD` equality, candidate SHA shape, repository identity, and a full byte comparison of `tools/k6-proofs`, `PROOFS` and `.github` against the approved commit. This hashes working-tree bytes with `git hash-object --no-filters`; `git status` is not used because the index makes `assume-unchanged`/`skip-worktree` able to hide a modified shared import.
2. **immutable snapshot** — all three trees are extracted with `git archive` into a private temp directory and the runner `exec`s that extract's own `run-proofs.sh`, before any credential is loaded. The handoff is authenticated, not trusted: the executing script must be the snapshot's own runner, the snapshot must be a distinct tree outside the checkout, and it must carry an unguessable ownership sentinel. Nothing is symlinked back to the checkout, because static rows read the proof corpus during k6 execution.
3. **catalog preflight** — validators run from the snapshot under `env -i` with a minimal allowlist.
4. **row selection and contract binding** — `all` expansion and the per-row manifest/scenario digest freeze. A selected row with no manifest at the approved ref fails closed rather than being silently skipped.

Credential, runtime-stamp and session resolution all happen only inside the verified runner.

### Receipts

- `harness-provenance.json` at the artifact root before the first row fires: matrix id, docs ref, repository identity, candidate SHA, runtime identity receipt, runner script digest, row selection with per-row manifest/scenario digests, start time. An immutable per-matrix copy is kept at `harness-provenance/<matrixId>.json`.
- every `runner-metadata.json` gains `docsRef`, `repository`, `matrixId`, `manifestPath`/`manifestSha256`, `scenarioPath`/`scenarioSha256`.
- every run directory gains `row-scenario.js` — the exact scenario source that fired, beside the copied `row-manifest.json`.
- both candidate-envelope validation layers bind those values fail-closed, with an exact key set, basename-only artifact references, and `authoritativeReceipt` permitted only on rows that have a row-scoped resolver.

Only validated values reach any public receipt; rejected operator input is recorded as `malformed`, and the snapshot/checkout/artifact-root/`$HOME` paths are scrubbed from any captured text.

Review boundaries are unchanged: `candidateOnly:true`, `foldRequiresReview:true`, `canonicalFoldForbidden:true`, `behaviorProof:false`.

## CI

`project81-k6-proof.yml` pins `actions/checkout` to `github.sha`, passes `--docs-ref "$DOCS_REF"` and `OPENCLAW_PROOFS_DOCS_REPOSITORY`, and gives the validators an explicit `--repo-root`.

## Validation

```
node --test tools/k6-proofs/scripts/__tests__/*.test.mjs   263 -> 318 pass / 0 fail
node --test tools/k6-proofs/tests/*.test.mjs                31 -> 31  pass / 0 fail
```

`bash -n` clean; `shellcheck -S warning` reports only two pre-existing SC2155 warnings; `git diff --check` clean.

Both defects were reproduced against the base commit rather than assumed: the pre-change validator still ENOENT-crashes from `tools/k6-proofs` where the new one matches the repository-root run exactly, and firing with the stale `25d330ef` ref now exits 78 with `{"check":"head-equals-docs-ref","rowsExecuted":0}` and no row executed.

The branch went through seven rounds of adversarial review; 22 real defects were found and fixed, each pinned by a regression test — including one dangerous bug introduced by this work itself (a rejected snapshot handoff recursively deleted the directory it claimed) which is now covered by a test asserting an unrelated bystander directory survives intact. The full table is in `output.md`.

## Agreed follow-ups (not blockers)

1. Committed symlink / gitlink / executable-mode comparison (content is compared; mode is not).
2. `analysis/project86-proof-issue-plan.corrected.json` still emits `--live` commands without `--docs-ref`; they now fail closed until `OPENCLAW_PROOFS_DOCS_REF` is exported. Left alone here because that file is guarded by its own idempotence and lock-serialization tests and touching it would collide with the separate R-CD-2 runtime-selection correction this branch excludes.
3. Inherited-token trust boundary: a token already exported by the caller is inherited by the bootstrap process by construction.

## Definition of done

A future Project 86 matrix cannot run from stale or ambiguous harness bytes, cannot double-resolve its catalog root, and carries enough immutable identity to prove which manifest/scenario contract produced every receipt.

### DIFF
--- a/.github/workflows/project81-k6-proof.yml
+++ b/.github/workflows/project81-k6-proof.yml
@@ -86,10 +86,14 @@ jobs:
       OPENCLAW_CREATE_DISPOSABLE_SESSIONS: ${{ inputs.create_disposable_sessions }}
       OPENCLAW_ALT_MODEL: ${{ inputs.openclaw_alt_model }}
       OPENCLAW_PROOFS_K6_OTLP_ENDPOINT: ${{ inputs.metrics_otlp_endpoint }}
+      OPENCLAW_PROOFS_DOCS_REPOSITORY: ${{ github.repository }}
+      DOCS_REF: ${{ github.sha }}
       K6_PROOF_OUT_DIR: ${{ github.workspace }}/project81-k6-proof-artifacts
     steps:
       - name: Checkout proof catalog
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.sha }}
 
       - name: Validate catalog and resolve rows
         id: rows
@@ -110,9 +114,11 @@ jobs:
           fi
           node --check tools/k6-proofs/scripts/list-runnable-rows.mjs
           node --check tools/k6-proofs/scripts/summarize-review-debt.mjs
-          node tools/k6-proofs/scripts/check-manifest-scenarios.mjs
-          node tools/k6-proofs/scripts/check-scenario-alignment.mjs
-          node tools/k6-proofs/scripts/check-proof-row-manifests.mjs
+          # One repository-root contract (#495): these produce identical results
+          # from the repository root and from tools/k6-proofs.
+          node tools/k6-proofs/scripts/check-manifest-scenarios.mjs --repo-root "$GITHUB_WORKSPACE"
+          node tools/k6-proofs/scripts/check-scenario-alignment.mjs --repo-root "$GITHUB_WORKSPACE"
+          node tools/k6-proofs/scripts/check-proof-row-manifests.mjs --repo-root "$GITHUB_WORKSPACE"
           if [[ "$ROWS_INPUT" == "live-suite" ]]; then
             rows="$(node tools/k6-proofs/scripts/list-runnable-rows.mjs --live-suite)"
           elif [[ "$ROWS_INPUT" == "all" ]]; then
@@ -144,7 +150,10 @@ jobs:
           if [[ "${{ inputs.metrics_push }}" != "true" ]]; then
             unset OPENCLAW_PROOFS_K6_OTLP_ENDPOINT
           fi
-          ./scripts/run-proofs.sh "$mode" "${{ steps.rows.outputs.rows }}" "${OPENCLAW_CANDIDATE_SHA:-}"
+          # Bind the matrix to the exact immutable docs/harness commit this job
+          # checked out (#496). The runner freezes it before any row fires and
+          # fails closed as harness infrastructure if the tree is stale or dirty.
+          ./scripts/run-proofs.sh "$mode" --docs-ref "$DOCS_REF" "${{ steps.rows.outputs.rows }}" "${OPENCLAW_CANDIDATE_SHA:-}"
 
       - name: Summarize review debt
         if: ${{ always() && !inputs.dry_run }}
--- a/RUNBOOKS/project-81/EXECUTABLE-SUITE.md
+++ b/RUNBOOKS/project-81/EXECUTABLE-SUITE.md
@@ -95,7 +95,8 @@ preflight (`check-manifest-scenarios`, `check-scenario-alignment`,
 Artifacts are written under `${K6_PROOF_OUT_DIR:-/tmp/k6-proof-runs}`:
 
 ```text
-<out-dir>/harness-provenance.json
+<out-dir>/harness-provenance.json              # newest matrix
+<out-dir>/harness-provenance/<matrix-id>.json  # immutable per-matrix copy
 <out-dir>/<candidate-sha>/<ROW>/<seat>/<timestamp-row>/
 ```
 
--- a/RUNBOOKS/project-81/EXECUTABLE-SUITE.md
+++ b/RUNBOOKS/project-81/EXECUTABLE-SUITE.md
@@ -78,16 +78,33 @@ Use disposable sessions unless you are deliberately proving behavior on a named 
 ```bash
 export OPENCLAW_CREATE_DISPOSABLE_SESSION=true
 export OPENCLAW_CREATE_DISPOSABLE_SESSIONS=true
-./scripts/run-proofs.sh --live "$ROWS"
+export OPENCLAW_PROOFS_DOCS_REF="$(git rev-parse HEAD)"
+./scripts/run-proofs.sh --live --docs-ref "$OPENCLAW_PROOFS_DOCS_REF" "$ROWS"
 ```
+
+A live run is bound to one immutable docs/harness commit (#496). `--docs-ref`
+(or `OPENCLAW_PROOFS_DOCS_REF`) is required and frozen at startup; the runner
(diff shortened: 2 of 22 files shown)


scribe-dandelion-cult:
Final scribe acceptance is bound to exact head `49585dd4dac2d8b7bfdd1d4bdf5d4c0ba430c32d`.

Independent review found and this normal descendant closed the remaining vacuous envelope-consumer coverage: canonical acceptance is now discriminating, observability mismatches are isolated one field at a time, and pending-debt fallbacks are named truthfully. Exact validation: proof-script suite **319/319**, root proof suite **31/31**, `git diff --check` clean. The immutable approved-ref execution, snapshot ownership, token isolation, fail-closed provenance, and catalog-root contracts have no remaining significant finding.

Accepted for normal merge; `PROOFS/**` remains untouched.

PR #510, opened 2026-08-12
fix(k6-proofs): WO-1217 proof harness authority repair

karmafeast:
## Summary

Repairs invalid continuation proof-row authority in the k6 harness (docs/harness only; no OpenClaw product change).

### R-CD-4
- **No OpenClaw behavior regression** in the prior failing runs: delegate spawned, exact debug log recorded `[continuation:targeted-return] Delivered to <target> from <child>`, diagnostics-otel present.
- Silent-wake delivery authority is now the shared post-run collector over the payload-free gateway journal line, not transcript `session.message` / `sessions.get` `TARGET-RECEIVED` text.
- PASS requires exactly one in-window target receipt and zero matching parent receipts, with independent child-identity gates.
- Public receipt publishes fingerprints/closed fields only.

### Product trace bug
- The product trace-parent bug remains tracked in **karmaterminal/openclaw#1251**. This PR does not compensate for it in behavioral assertions.

### R-CD-CHAINED-DEPTH-2
- Nested child→grandchild call now explicitly requests `fanoutMode="tree"`; outer parent→child is unchanged.
- Requires two distinct nonce-bound hop identities.
- Uses the **same** shared targeted-return collector for grandchild→root routing (tree multi-target / intermediate ancestors allowed).
- Removed the fictional three-subtest claim from the manifest.

### R-CD-MODEL-TOOL
- Requires disposable parent.
- Pre/post `sessions.list { spawnedBy: <parent>, limit: 100 }` set-diff; exactly one new child; model byte from that row.
- Parent `sessions.get` must show a nonce-bound **return** via the exact `MODEL-TOOL-CHILD <nonce> MODEL <provider/model>` marker.
- Paraphrased schedule acknowledgements and generic nonce-bearing assistant text are **not** return authority (review fix for review 4920192489).
- Child self-report is auxiliary and cannot establish model equality.
- No direct production DB query as PASS authority.

### Tempo search IDs
- Accept exactly 31 hex characters by left-padding one `0`.
- Fetch and verify decoded nonzero payload trace/span IDs.
- Reject other malformed lengths and all-zero IDs.

### Historical artifacts
- **Old PROOFS artifacts remain unchanged.** `PROOFS/INDEX.json` was not repointed.
- Replay of redacted journals remains unproven (no manufactured PASS).

## Review follow-up
- Tightened `parentReturnContainsNonce` to require the exact child-return marker.
- Added adversarial case: `Scheduled delegate for nonce-x; waiting for completion.` => `false`.
- Code fix commit: `dffb1ceb8c78c789442c18135f00618b564da85d`.

## Tests
```text
node --test tools/k6-proofs/tests/r-cd-model-tool-authority.test.mjs
# 5 pass / 0 fail

node --test tools/k6-proofs/tests/**/*.test.mjs tools/k6-proofs/scripts/__tests__/**/*.test.mjs
# 363 pass / 1 fail (pre-existing: candidate envelope corpus INDEX schema drift on current_sha proofs-manifest.v1 vs k6.proofs-manifest.v1; unrelated to this harness authority change)
node tools/k6-proofs/scripts/check-manifest-scenarios.mjs --repo-root "$PWD"  # pass
node tools/k6-proofs/scripts/check-scenario-alignment.mjs --repo-root "$PWD"  # pass
node tools/k6-proofs/scripts/check-proof-row-manifests.mjs --repo-root "$PWD" # pass
```

## Remaining live-proof gap
Exact live re-runs still required on prince seats with disposable sessions to mint fresh PASS-candidate artifacts under the new authority for:
- R-CD-4
- R-CD-CHAINED-DEPTH-2 (with nested `fanoutMode=tree`)
- R-CD-MODEL-TOOL (disposable parent + spawnedBy set-diff + exact child-return marker)

Do not fold historical PARTIAL runs as PASS under the new detector without a new live fire.

## Base
- docs `origin/main` @ `513a74ad1db3cfbb5ac10cc45155ee1e1acb911a`

### DIFF
--- a/tools/k6-proofs/lib/r-cd-model-tool-authority.mjs
+++ b/tools/k6-proofs/lib/r-cd-model-tool-authority.mjs
@@ -45,35 +45,170 @@ export function resolveModelToolChildAuthority({
   postKeys,
   sessionsPayload,
   requestedModel,
+  eventChildSessionKey,
+  parentSessionKey,
 }) {
   const diff = diffSpawnedByChildren(preKeys, postKeys);
+  const sessions = Array.isArray(sessionsPayload?.sessions)
+    ? sessionsPayload.sessions
+    : (Array.isArray(sessionsPayload) ? sessionsPayload : []);
+  const pre = new Set(Array.isArray(preKeys) ? preKeys : []);
+  const eventKey = typeof eventChildSessionKey === 'string' && eventChildSessionKey.length > 0
+    ? eventChildSessionKey
+    : null;
   if (!diff.uniqueNewChildKey) {
     return {
       ...diff,
       childSessionKey: null,
+      childSession: null,
       childMetadataModelByte: null,
       modelMatches: false,
-      failureCategory: diff.ambiguous
-        ? 'multiple-new-children'
-        : 'zero-new-children',
+      authoritySource: null,
+      failureCategory: diff.ambiguous ? 'multiple-new-children' : 'zero-new-children',
     };
   }
+  if (eventKey && (pre.has(eventKey) || eventKey !== diff.uniqueNewChildKey)) {
+    const diffChild = sessions.find((session) => session?.key === diff.uniqueNewChildKey) || null;
+    return {
+      ...diff,
+      childSessionKey: diff.uniqueNewChildKey,
+      childSession: diffChild,
+      childMetadataModelByte: modelFromSessionMetadata(diffChild),
+      modelMatches: false,
+      authoritySource: null,
+      failureCategory: pre.has(eventKey)
+        ? 'event-child-preexisting'
+        : 'event-child-diff-mismatch',
+    };
+  }
+  const selectedKey = diff.uniqueNewChildKey;
 
-  const sessions = Array.isArray(sessionsPayload?.sessions)
-    ? sessionsPayload.sessions
-    : (Array.isArray(sessionsPayload) ? sessionsPayload : []);
-  const child = sessions.find((session) => session?.key === diff.uniqueNewChildKey) || null;
+  const child = sessions.find((session) => (
+    session?.key === selectedKey
+    && (!parentSessionKey
+      || session.spawnedBy === parentSessionKey
+      || session.parentSessionKey === parentSessionKey)
+  )) || null;
+  if (!child) {
+    return {
+      ...diff,
+      childSessionKey: null,
+      childSession: null,
+      childMetadataModelByte: null,
+      modelMatches: false,
+      authoritySource: null,
+      failureCategory: 'child-parent-mismatch',
+    };
+  }
   const childMetadataModelByte = modelFromSessionMetadata(child);
   const requested = normalizeModel(requestedModel);
   const modelMatches = !!childMetadataModelByte && childMetadataModelByte === requested;
   return {
     ...diff,
-    childSessionKey: diff.uniqueNewChildKey,
+    childSessionKey: selectedKey,
+    childSession: child,
     childMetadataModelByte,
     modelMatches,
+    authoritySource: eventKey
+      ? 'event-correlated-sessions-list'
+      : 'spawned-by-set-diff',
     failureCategory: !childMetadataModelByte
       ? 'missing-child-model-byte'
       : (modelMatches ? null : 'model-mismatch'),
+  };
+}
+
+export function mergeModelToolChildAuthorityEvidence(currentEvidence, authority) {
+  const current = currentEvidence || {};
+  if (
+    current.child_identity_conflict === true
+    || current.spawned_by_diff_ambiguous === true
+  ) {
+    return { ...current };
+  }
+  if (authority?.ambiguous === true) {
+    return {
+      ...current,
+      spawned_by_diff_ambiguous: true,
+      model_matches: false,
+      model_classification_reason: 'spawnedBy set-diff produced multiple new children',
+    };
+  }
+  const authorityChildKey = typeof authority?.childSessionKey === 'string'
+    && authority.childSessionKey.length > 0
+    ? authority.childSessionKey
+    : null;
+  if (
+    authorityChildKey
+    && (
+      (
+        typeof current.child_session_key === 'string'
+        && current.child_session_key.length > 0
+        && current.child_session_key !== authorityChildKey
+      )
+      || (
+        typeof current.event_child_session_key === 'string'
+        && current.event_child_session_key.length > 0
+        && current.event_child_session_key !== authorityChildKey
+      )
+    )
+  ) {
+    return {
+      ...current,
+      child_identity_conflict: true,
+      spawned_by_diff_ambiguous: true,
+      model_matches: false,
+      model_classification_reason: 'conflicting authoritative child-session keys',
+    };
+  }
+  const hasStickyAuthority = (
+    current.child_session_metadata_observed === true
+    && typeof current.child_session_key === 'string'
+    && current.child_session_key.length > 0
+    && typeof current.child_metadata_model_byte === 'string'
+    && current.child_metadata_model_byte.length > 0
+  );
+  const hasNewAuthority = (
+    typeof authority?.childSessionKey === 'string'
+    && authority.childSessionKey.length > 0
+    && typeof authority.childMetadataModelByte === 'string'
+    && authority.childMetadataModelByte.length > 0
+  );
+
+  if (!hasNewAuthority) {
+    if (hasStickyAuthority) return { ...current };
+    return {
+      ...current,
+      spawned_by_diff_empty: authority?.empty === true,
+      spawned_by_diff_ambiguous: authority?.ambiguous === true,
+      model_classification_reason: authority?.failureCategory || current.model_classification_reason,
+    };
+  }
+
+  const child = authority.childSession || {};
+  return {
+    ...current,
+    child_session_key: authority.childSessionKey,
+    child_session_observed: true,
+    child_session_metadata_observed: true,
+    child_metadata_model_byte: authority.childMetadataModelByte,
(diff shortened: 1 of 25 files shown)


tools/k6-proofs/lib/r-cd-model-tool-authority.mjs
```diff
@@ -0,0 +1,122 @@
...
+  for (const message of Array.isArray(messages) ? messages : []) {
+    const role = message?.role;
+    if (role !== 'user' && role !== 'assistant' && role !== 'system') continue;
+    const text = messageText(message);
+    if (typeof text !== 'string' || !text.includes(nonce)) continue;
+    if (text.includes('[k6-proof-harness]')) continue;
+    // Parent schedule ack is not child→parent return authority.
+    if (/\bMODEL-TOOL-PARENT-SCHEDULED\b/.test(text)) continue;
+    // Accept child return marker or other non-schedule nonce-bearing return text.
+    if (/\bMODEL-TOOL-CHILD\b/.test(text) || role === 'system' || role === 'assistant') {
```
scribe-dandelion-cult: **P1 — generic assistant nonce text can falsely prove child return.** This branch accepts any assistant message containing the nonce, not only the required `MODEL-TOOL-CHILD <nonce> MODEL ...` completion payload. Reproduction: `parentReturnContainsNonce([{role:'assistant', content:'Scheduled delegate for nonce-x; waiting for completion.'}], 'nonce-x')` returns `true`. A model that paraphrases the schedule acknowledgement therefore satisfies `parent_return_nonce_bound` even if completion delivery never lands. Require the exact child-return marker (or a separately typed structured system-return contract), and add this adversarial schedule-paraphrase case alongside the existing exact-marker test.

scribe-dandelion-cult (changes requested):
One false-PASS authority bug remains in R-CD-MODEL-TOOL. The shared targeted-return and Tempo changes otherwise look coherent from this pass. Please tighten parent-return authority and add the adversarial case before live reruns.

tools/k6-proofs/lib/targeted-return-receipt.mjs
```diff
@@ -0,0 +1,221 @@
...
+    if (receipt.authority !== 'gateway-journal-targeted-return') {
+      return { valid: false, reason: 'invalid-authority' };
+    }
+    if (receipt.candidateOnly !== true || receipt.foldRequiresReview !== true) {
+      return { valid: false, reason: 'review-flags' };
+    }
+    if (receipt.verdict !== 'PASS-candidate' && receipt.verdict !== 'PARTIAL-candidate') {
+      return { valid: false, reason: 'invalid-verdict' };
+    }
+    if (receipt.verdict === 'PASS-candidate') {
```
scribe-dandelion-cult, line 266: **P1 — this `authoritativeReceipt` is unsigned and accepts structurally impossible PASS.** The existing R-CD-2/R-CD-TOKEN authoritative receipts HMAC-bind resolver output to `OPENCLAW_GATEWAY_TOKEN`; this validator ignores the token argument entirely. A hand-written receipt with `verdict: PASS-candidate`, valid-looking counts/fingerprints, and `structuralOk:false` returns `{valid:true}` both with and without a token. The run-result SHA only rehashes substituted content, so it is not an independent seal. Use the established `createHmac`/`timingSafeEqual` pattern and require `structuralOk === true` for PASS before exposing this through the same authoritative-receipt contract.

tools/k6-proofs/scenarios/r-cd-model-tool.js
```diff
@@ -83,11 +95,81 @@ export default function() {
...
       socket.setTimeout(() => {
-        // Deliberately do NOT include requestedModel in the child task. The child
-        // must report its own runtime identity instead of echoing the request.
+        if (!evidence.pre_spawned_by_captured) {
+          console.error('✗ pre-dispatch sessions.list {spawnedBy} baseline missing');
+          failures.add(1);
+          socket.close();
+        }
+      }, 5000);
+      socket.setTimeout(() => {
```
scribe-dandelion-cult: **P1 — dispatch is silently lost when the pre-baseline RPC takes >500 ms.** `sessions.list` starts at +300 ms, but this sole dispatch timer fires at +800 ms and returns permanently if the response has not arrived. If the baseline arrives between 800 ms and the 5 s watchdog, the watchdog sees it, all post polls no-op because dispatch was never accepted, and the row idles for 180 s as a latency-dependent false FAIL. Trigger dispatch from the successful `pre` response (once), or explicitly re-arm it until the watchdog; add a delayed-baseline test.

scribe-dandelion-cult (changes requested):
The first false-PASS finding is fixed. Two independent acceptance blockers remain: the new authoritative receipt silently downgrades integrity compared with existing authoritative rows, and model-tool can drop its only dispatch when the baseline RPC is slower than 500 ms.

karmafeast:
## Review fix (4920347189) @ `f7e307d7`

### 1) Targeted-return authoritative receipt HMAC + structuralOk
- Receipts now seal closed fields with `hmac-sha256-gateway-token-v1` via `createHmac`/`timingSafeEqual` (same contract as R-CD-2/R-CD-TOKEN).
- Collector requires `OPENCLAW_GATEWAY_TOKEN`, self-validates, writes public-safe sealed JSON only.
- `validateTargetedReturnReceipt(receipt, signingKey, row)` rejects missing token, unsigned forged PASS, tampered fields, wrong key.
- **PASS requires `structuralOk === true`** (hand-forged PASS with structuralOk:false is rejected even if HMAC matches that impossible body).
- Candidate contract/validator + `run-proofs.sh` plumb the token at collector/validator boundaries.

### 2) R-CD-MODEL-TOOL baseline-gated dispatch
- Removed fixed +800ms sole dispatch timer (lost dispatch when pre `sessions.list` >500ms).
- `createModelToolDispatchGate()` dispatches **exactly once** from the successful pre-baseline response.
- 5s watchdog still fail-closes if baseline never arrives; late baseline after fail-closed cannot resurrect dispatch.
- Delayed-baseline regression tests cover once-only + never-before-baseline.

### Validation
- Focused: 36 pass / 0 fail
- Full k6-proofs: 367 pass / 1 fail (pre-existing INDEX schema drift)
- Manifest checks: pass
- Historical PROOFS untouched; do not merge.

scribe-dandelion-cult (approved):
Re-review at exact head `f7e307d7ab51fcfd10480ff112d54ac2cd2d31b0`: both changes-requested rounds are resolved.

- exact child-return marker closes the schedule-paraphrase false PASS;
- targeted-return receipt is HMAC-sealed with the gateway token and PASS requires `structuralOk=true`;
- model dispatch is armed exactly once by the successful pre-baseline response, with delayed-baseline/watchdog coverage;
- independent focused authority suite: 20/20 pass;
- independent full harness suite: 367/368, with the sole INDEX/manifest schema failure on unchanged pre-existing corpus bytes.

Code/harness review approved. This does **not** merge the PR or clear the separately required exact live reruns for R-CD-4, R-CD-CHAINED-DEPTH-2, and R-CD-MODEL-TOOL after the product trace fix.

scribe-dandelion-cult:
Exact-baseline classification receipt (new policy): on PR base `513a74ad1db3cfbb5ac10cc45155ee1e1acb911a`, `node --test tools/k6-proofs/scripts/__tests__/candidate-run-result.test.mjs` is **28/29**, failing the same `candidate envelope is outside and invisible to canonical corpus validation` assertion because bound `PROOFS/a7ef0317.../proofs-manifest.json` has schema `openclaw.k6.proofs-manifest.v1` while the canonical validator expects `openclaw.proofs.manifest.v1` plus `capture_sha` and `rows[]`. This exactly reproduces the sole 367/368 head failure on the untouched base. Classification: pre-existing INDEX/manifest corpus drift, not introduced by `f7e307d7`. PR remains approved/unmerged; live R-CD-4/chained/model-tool reruns remain required.

karmafeast:
Observation-repair follow-up pushed at exact head `10f98e60cd48eba4d598a0e8805ec42d632a1326` (docs/harness only).

- `R-CD-MODEL-TOOL`: request-phase correlation for overlapping `sessions.list` calls; exact unique pre/post `spawnedBy` child remains mandatory; nonce-bound event identities may only confirm it; positive authority survives empty polls, while replacement children, multiple children, or conflicting events fail closed; PASS waits for a hard 30-second post-dispatch authority confirmation.
- `R-CD-CHAINED-DEPTH-2`: disposable root required; child and grandchild identities resolve from repeated `sessions.list { spawnedBy }` ancestry rather than truncated task titles; ambiguity is sticky; PASS requires two confirmations plus a >=30-second grandchild stability receipt.
- Signed targeted-return collector independently requires non-ambiguous ancestry, both ancestry authority sources, confirmation counts, and the >=30-second timestamp interval.

Validation: focused authority/collector suite `40/40`; full harness `378/379`. The sole red is the previously proven base/head-identical candidate-envelope `INDEX`/manifest schema mismatch; none is attributable to this repair. Final independent `/code-review`: no significant issues.

Historical artifacts and `PROOFS/INDEX.json` remain untouched. Fresh exact-product live fires are still required before proof credit.

scribe-dandelion-cult (approved):
Exact-head re-review at 10f98e60cd48eba4d598a0e8805ec42d632a1326: the observation-authority follow-up closes the overlapping-RPC, replacement-child, event-conflict, truncated-ancestry, early-close, and configurable-stability false-PASS paths. Focused authority/collector validation is 40/40; full harness is 378/379 with the sole failure exactly reproduced on the base as the existing INDEX/manifest schema mismatch. Independent final /code-review found no significant issues. APPROVE docs/harness code correctness at this exact head. This does not merge the PR or grant historical/live proof credit; fresh exact-product live fires remain mandatory.

Issue #520, opened 2026-08-27
R-CD-CHAINED-DEPTH-2: deduplicate aliased task detail fields before marker cardinality


scribe-dandelion-cult:
## Problem

The public `tasks.get` projection for a subagent exposes its progress summary in both `progressSummary` and `result`. R-CD-CHAINED-DEPTH-2 concatenated `progressSummary`, `terminalSummary`, and `result` before enforcing exactly-once terminal markers. Byte-identical aliases therefore counted `CHILD-DONE` or `GRANDCHILD-DONE` twice and rejected a valid completed two-task ledger.

Exact run [`33028327545`](https://github.com/karmaterminal/karmaterminal-openclaw-docs/actions/runs/33028327545), artifact `9629388466`, against runtime `5d79066ea4926c69fcf7ac8376688a0babbf7797` produced two distinct succeeded/delivered tasks with root -> child -> grandchild ownership. The depth-1 row had `lastToolName=continue_work`, one final `CHILD-DONE`, and the depth-2 row had one final `GRANDCHILD-DONE`. The observer remained PARTIAL because the same public progress byte was projected into both fields.

This is a harness projection/cardinality defect, not a product behavior defect.

## Required behavior

- Deduplicate byte-identical public detail fields before marker cardinality checks.
- Do not globally relax exactly-once marker enforcement.
- A marker repeated within one field, or repeated across genuinely different field values, must remain non-PASS.
- Preserve exact two-task/two-run identity, root -> child -> grandchild ownership, completed/delivered state, post-dispatch timing, full-prompt nonce binding, `lastToolName=continue_work`, and final marker requirements.
- Repeated use of the same high-entropy row nonce in task instructions is valid; stale/wrong nonce and mixed-row records remain invalid.

## Existing successor to review

`codeagent/129388-final-authority-harness-cure@08fc038d998baae83391f24594a18890e4db1281` deduplicates identical detail strings and includes a sanitized exact-run projection fixture. Review/harden it rather than assuming it is complete.

## Acceptance

1. The exact public shape with identical `progressSummary` and `result` yields one marker occurrence and a valid ledger.
2. Duplicate marker text within a single field remains rejected.
3. Distinct fields that each independently contain the marker remain rejected unless they are byte-identical aliases.
4. Wrong ownership, duplicate task/run identity, stale/pre-dispatch rows, missing delivery, wrong nonce, wrong depth, or missing final marker remain non-PASS.
[...]

Issue #519, opened 2026-08-27
R-CD-TOKEN: make public cursor recovery dispatch deterministic across subscription race


scribe-dandelion-cult:
## Problem

R-CD-TOKEN subscribes to the disposable origin, resolves the exact origin/delegate task lineage, snapshots the initial-origin transcript cursor with `sessions.get`, and binds the normal assistant return by exact announce run ID and post-cursor sequence.

The scenario initiated the first cursor request through `socket.setTimeout(..., 0)`. Under k6 WebSocket execution in exact run [`33028327545`](https://github.com/karmaterminal/karmaterminal-openclaw-docs/actions/runs/33028327545), artifact `9629388466`, that zero-delay timer remained armed and no `sessions.get` request was sent. Both origin and delegate tasks completed, but `origin_cursor_snapshot_accepted=false`, no return could bind, and the row expired PARTIAL.

A related subscription race exists when the public assistant return is durably projected before the live origin subscription is accepted. `sessions.get` must recover the exact post-origin-cursor assistant message without consulting private logs.

This is a deterministic harness scheduling/observer defect, not a product behavior defect.

## Required behavior

- Issue the first `sessions.get` synchronously as soon as both subscription acceptance and exact task-ledger identity are ready.
- Use positive-delay timers only for subsequent recovery polls.
- Recover the normal return from the public transcript when it predates subscription acceptance.
- Bind only the disposable origin session, exact accepted delegate child/run announce ID, exact nonce sentinel once, and message sequence newer than the initial-origin cursor.
- Deduplicate live and transcript observations by message sequence.
- A second distinct matching return, wrong session/run/nonce, stale cursor, root-only result, private-log-only text, missing public event, rejected snapshot, or buffer overflow remains non-PASS.

## Existing successor to review

`codeagent/129388-final-authority-harness-cure@08fc038d998baae83391f24594a18890e4db1281` replaces the zero-delay initiation with a synchronous request and adds transcript recovery. Review/harden it, especially with a deterministic scheduler/RequestTracker test that proves `sessions.get` is actually emitted.

## Acceptance

1. Deterministic scenario-level test proves the first `sessions.get` is sent after subscription + lineage readiness without advancing a zero-delay timer.
2. Frozen pre-subscription public return is rejected by the live parser and accepted by transcript recovery.
3. Live and transcript delivery of the same message sequence counts once; a second sequence fails duplicate authority.
[...]

PR #522, opened 2026-08-27
fix(proofs): close R-CD-2, depth, and token observer debt

scribe-dandelion-cult:
## Summary

Hardens the existing `08fc038d` candidate remedies for three proof-harness defects. HMAC receipts remain the sole candidate verdict authority. Fail-closed topology, nonce, and marker contracts are not relaxed.

- Closes #519 — R-CD-2 public-safe per-gate diagnostics with table-driven inversion coverage.
- Closes #520 — byte-identical `progressSummary`/`result` aliases count once; intra-field and distinct-field repeats stay non-PASS.
- Closes #521 — first `sessions.get` is synchronous after subscription + lineage; transcript recovers the pre-subscription public return; live+transcript copies of one sequence dedupe.

### #519 / R-CD-2
Public resolver stdout now uses `rCd2PublicResolutionOutput`: boolean lifecycle, topology, and join gates only. HMAC-signed receipt stays the sole verdict. Inverting any one gate keeps the row non-PASS while unrelated gates retain their expected values. Unique nonce/reason binding when `sessions.send` omits `traceId` is preserved. Resolution JSON and receipt pass sanitizer/public-safety checks.

### #520 / R-CD-CHAINED-DEPTH-2
Existing Set-dedup of identical public detail strings is now covered: identical `progressSummary`/`result` aliases form a valid two-task ledger; a repeated marker inside one field is rejected; distinct non-identical fields that each contain the marker are rejected. Ownership, completed/delivered, timing, nonce, `continue_work`, and final-marker contracts are unchanged.

### #521 / R-CD-TOKEN
`tokenOriginCursorSnapshotDispatch` emits the first `sessions.get` immediately after subscription + exact origin lineage, without a zero-delay timer. Recovery polls require a positive delay. Live parser still rejects a pre-subscription public assistant message; transcript recovery accepts it. Live and transcript copies of one `messageSeq` count once; a second sequence is PARTIAL. Wrong session/run/nonce/cursor/root/private-log/missing-event negatives remain.

## Validation
- `node --test` over `tools/k6-proofs/scripts/__tests__` and `tools/k6-proofs/tests`: **524 pass / 0 fail**
- `node --check` on changed executable scripts/scenarios
- manifest/scenario alignment, row-manifest, telemetry contract, corpus `--current`, sanitizer, candidate-envelope, workflow/matrix, and proof-closure surfaces covered by those suites/checks
- `git diff --check` clean against `08fc038d`

## Limits
Does not refire Project-81, mutate a gateway, or fold a live PASS. Product behavior is unchanged; this is harness/authority debt only.

### DIFF
--- a/tools/k6-proofs/lib/r-cd-2-authoritative-receipt.mjs
+++ b/tools/k6-proofs/lib/r-cd-2-authoritative-receipt.mjs
@@ -179,6 +179,15 @@ export function rCd2AuthorityChecks(evidence, correlation) {
   };
 }
 
+/** Public-safe resolver stdout. Booleans only; HMAC receipt remains verdict authority. */
+export function rCd2PublicResolutionOutput(receipt, evidence, correlation) {
+  return {
+    verdict: receipt?.verdict || null,
+    receipt: 'r-cd-2-authoritative-receipt.json',
+    checks: rCd2AuthorityChecks(evidence, correlation),
+  };
+}
+
 function categoryFor(evidence, correlation) {
   if (evidence?.channel_message_observed === true) return 'silent-channel-delivery';
   if (evidence?.failureCategory === 'delegate-replay-unsafe' &&
--- a/tools/k6-proofs/lib/r-cd-token-contract.js
+++ b/tools/k6-proofs/lib/r-cd-token-contract.js
@@ -245,6 +245,55 @@ export function tokenOriginCursorFromMessages(
 }
 
 /**
+ * Decide the first public cursor recovery action. The first sessions.get must
+ * fire synchronously once subscription and exact origin lineage are ready.
+ * Positive-delay timers are only for later recovery polls; delay <= 0 is idle.
+ */
+export function tokenOriginCursorSnapshotDispatch({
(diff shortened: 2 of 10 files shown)


scribe-dandelion-cult (commented):
Independent scribe review: CLEAN (GitHub cannot record APPROVE because the review identity also authored the branch).

The delta remains narrowly scoped to #519/#520/#521. R-CD-2 diagnostics expose booleans only and do not displace the HMAC receipt as verdict authority; all lifecycle/topology/join gates have inversion coverage. Depth deduplicates only byte-identical public aliases while preserving genuine duplicate-marker rejection. Token cursor initiation is synchronous at lineage readiness, positive-delay-only for recovery, and live/transcript observations deduplicate by message sequence while a second sequence remains non-PASS.

Independent direct run: 524/524 proof-harness tests pass; diff/check and public-safety boundaries are clean. PR body issue numbering was corrected before this review. GitNexus process mapping was unavailable because the registered docs index uses an incompatible newer LadybugDB storage version; no graph-derived claim is credited.

Issue #523, opened 2026-08-27
Project-81 readiness must inspect the supplied gateway target, not runner host config


scribe-dandelion-cult:
## Problem

Project-81 accepts an explicit `gateway_ws`, isolated seat name, gateway unit, expected depth, candidate SHA, and runtime SHA. Its selected-row readiness still derived continuation configuration from the self-hosted runner's host OpenClaw config instead of the supplied isolated gateway target.

Exact infrastructure-invalid run: https://github.com/karmaterminal/karmaterminal-openclaw-docs/actions/runs/33034338560

- docs head: `6e6dee501a06e6e02ecca034912459faf39b7ad0`
- runtime: `5d79066ea4926c69fcf7ac8376688a0babbf7797`
- artifact: `9631573008`
- artifact digest: `sha256:aa2f2d47bd904ca992f9575097613066462723d59d7f8d1ba993bb2354a5872e`
- rows: `R-CD-2,R-CD-CHAINED-DEPTH-2,R-CD-TOKEN`

The new isolated target on `127.0.0.1:19893` passed HTTP and WebSocket health/status, exact-model smoke, Tempo reachability, and an explicit seat-readiness preflight with configured/effective/required depth `5/5/2`. The workflow then failed before model/proof traffic with:

- class: `harness-infrastructure`
- stage: `seat-readiness`
- exit: `78`
- reason: `configured-depth-unknown`
- rows executed: `0`

No row signed receipt or candidate envelope exists because no row ran. This is not a product or scenario verdict.

## Required behavior

1. Runner-side readiness must inspect the gateway identified by the workflow's explicit `gateway_ws` and auth token, not ambient host config.
2. Configured and effective continuation depth must be derived from a public authenticated target-gateway RPC/readiness surface or a cryptographically/identity-bound preflight receipt for that exact target.
3. `expected_max_spawn_depth` is an assertion, never a substitute for observed configured/effective depth.
4. A host config with depth missing or `1` must not override a target gateway reporting depth `5`.
5. Conversely, a target gateway reporting insufficient/unknown depth must fail closed even when the host config is sufficient.
6. Preserve pre-dispatch safety: zero row/model traffic when target readiness cannot be established.
7. Do not expose tokens, private config, session keys, transcripts, or provider credentials in logs/artifacts.
8. Bind readiness to docs head, candidate/runtime SHA, gateway URL fingerprint, seat, unit, selected rows, and required/expected depth.

[...]

PR #524, opened 2026-08-27
Fix Project-81 target-bound seat readiness

scribe-dandelion-cult:
Fixes #523

## Summary
- read continuation config and effective spawn depth from authenticated `config.get` on the explicit `gateway_ws` target
- bind the public target observation to the credential-free gateway fingerprint, seat, unit, docs head, candidate/runtime SHA, selected rows, and required/expected depth
- preserve fail-closed zero-row infrastructure stops and existing live-run safety gates
- add deterministic A/B/C/D target-vs-host and binding controls

## Validation
- complete proof-harness Node suite: 478/478
- manifest/scenario/row/telemetry/corpus validators pass
- all 155 harness JavaScript files pass `node --check`

No live proof was run and no proof corpus content changed.

### DIFF
--- a/.github/workflows/project81-k6-proof.yml
+++ b/.github/workflows/project81-k6-proof.yml
@@ -213,8 +213,8 @@ jobs:
         shell: bash
         run: |
           set -euo pipefail
-          # The runner executes ON the prince's own host, so the authoritative
-          # token for THIS seat's gateway is the one in that seat's own config.
+          # Host config is an authentication source only. Target readiness and
+          # continuation depth are always read from inputs.gateway_ws by RPC.
           # A single shared secret cannot authenticate against six gateways that
           # each hold a different token -- it only ever matched one seat, and
           # every other seat failed at `connect`, surfacing later and confusingly
--- a/tools/k6-proofs/docs/PROOF-RUN-METHOD.md
+++ b/tools/k6-proofs/docs/PROOF-RUN-METHOD.md
@@ -109,6 +109,12 @@ The live helper obtains continuation config through authenticated `config.get`
 on `OPENCLAW_GATEWAY_WS`; host config can provide authentication but never the
 target depth. Its public observation binds only a credential-free gateway
 fingerprint plus seat, unit, docs, candidate/runtime, rows, and depth contract.
+Candidate and runtime are separate 40-character identities and may differ at
+this readiness layer. The runner records that state as
+`candidateMatchesRuntime:false`; readiness does not promote it to exact-target
+evidence. Rows that require exact execution, including R-CD-TOKEN and
+R-CD-CHAINED-DEPTH-2, retain their row-specific equal-build gates, and the
+candidate-envelope validator withholds exact PASS authority on mismatch.
 
 ### 4. Dry-run the selected set
 
--- a/tools/k6-proofs/docs/PROOF-RUN-METHOD.md
+++ b/tools/k6-proofs/docs/PROOF-RUN-METHOD.md
@@ -89,6 +89,7 @@ OPENCLAW_CANDIDATE_SHA=<40-char-sha> \
 OPENCLAW_SEAT_NAME=<seat> \
 OPENCLAW_SESSION_KEY=<scratch-or-disposable-session> \
 OPENCLAW_GATEWAY_TOKEN=*** \
+OPENCLAW_GATEWAY_WS=<target-gateway-ws-url> \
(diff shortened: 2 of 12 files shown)


scribe-dandelion-cult:
**SCRIBE REVIEW — needs one contract resolution before merge**

`targetReadinessBindingErrors(..., { requireComplete: true })` now rejects every live run where `candidateSha !== runtimeBuildSha`. That is broader than issue #523 and conflicts with the existing runner contract at `run-proofs.sh`, which deliberately records both identities, warns on mismatch, and lets row-specific/candidate-envelope policy decide whether the evidence is exact, partial, or transposable.

This matters for composite execution: the deployed runtime identity can differ from a PR/presentation source identity while both must remain explicit. We must not globally collapse those axes in target *readiness*.

Please byte-walk current callers and resolve one of these shapes explicitly:

1. If distinct candidate/runtime identities remain supported, require both to be valid and bound but remove global equality from target readiness; add a deterministic valid-distinct-identity test and preserve downstream `candidateMatchesRuntime=false` / row-specific exactness gates.
2. If the repository has intentionally abolished distinct identities for all live Project-81 rows, make that contract consistent by updating policy/docs/tests and removing the contradictory warn-and-continue path. Do not choose this merely to keep the new helper green.

No other correctness finding in the reviewed delta. GitNexus is not credited because the registered docs index is unavailable under the current LadybugDB storage version.

scribe-dandelion-cult:
Resolved the scribe finding in `68ed52772cbfc3b4d223974c7fa84112ec1201f8`.

**Semantic decision:** distinct valid candidate/presentation and deployed-runtime identities remain supported. Target readiness now requires and binds both independent 40-character SHAs but no longer requires equality. Equality remains an authority concern downstream, not a target-config readiness concern.

**Code/test changes**
- Removed only the global `candidateSha === runtimeBuildSha` rejection from `targetReadinessBindingErrors(..., { requireComplete: true })`; all gateway/seat/unit/docs/candidate/runtime/row/depth binding and validity checks remain intact.
- Added a deterministic helper + real preflight regression using candidate `bbbb…` and runtime `cccc…`; `--require-target-binding` returns `PASS-candidate`, retains both identities byte-for-byte, and reports a valid binding.
- Documented the two-axis contract in `docs/PROOF-RUN-METHOD.md`.

**Axis classification proof**
- `run-proofs.sh` remains unchanged: mismatch warns/continues and writes `runtimeIdentity.candidateMatchesRuntime=false`.
- R-CD-TOKEN retains its pre-dispatch exact-build gate and its signed resolver test proves mismatch becomes `PARTIAL-candidate` and cannot become authority.
- R-CD-TOKEN and R-CD-CHAINED-DEPTH-2 candidate-envelope checks still reject runtime/candidate mismatch as `build identity mismatch`, withholding exact PASS authority.
- Corpus transposition semantics remain unchanged: carried source/runtime evidence does not become exact-target execution.
- Issue #523’s authenticated target-vs-host config binding and harness-infrastructure zero-row fail-closed behavior were not changed.

**Exact validation**
- Focused readiness/row-authority/envelope/zero-row suites: **85/85 passed**.
- Full proof harness: **532/532 passed**.
- Dedicated sanitizer/public-artifact-safety: **23/23 passed**.
- Runnable-row listing; manifest/scenario registry; scenario alignment; proof-row manifest; telemetry-contract; and current-corpus validators all passed. Current corpus: 41 rows = 38 required + 3 supplemental; 0 missing manifests.
- Node syntax, shell syntax, and `git diff --check` passed.

**Branch equality:** local = tracking = server = `68ed52772cbfc3b4d223974c7fa84112ec1201f8`.

scribe-dandelion-cult (commented):
Independent scribe re-review: CLEAN.

The prior identity-axis finding is resolved. Target readiness now validates and binds candidate/runtime as two independent 40-character identities without granting exactness. Existing `candidateMatchesRuntime=false`, row-specific equal-build signed authorities, and candidate-envelope rejection remain responsible for exact execution claims.

The authenticated target `config.get` fix, URL credential stripping/fingerprint, A/B/C/D fail-closed fixtures, zero-row infrastructure stop, and provenance propagation remain intact. Independent full suite: 532/532 pass; final diff/check is clean. GitNexus remains uncredited because the registered docs index uses an incompatible LadybugDB storage version.

GitHub cannot record APPROVE because the review identity also authored the branch; this CLEAN comment is the independent scribe verdict.

